PR #23, opened 2019-05-14
Fix calculation of mean_squared_log_error

akondas:
Instead of focusing on who copies the code from whom and how well the license is understood (https://github.com/php-ai/php-ml/issues/379), let's focus on providing the correct value.

Here is an example from sklearn package:
Screenshot from 2019-05-14 08-14-24

### DIFF
--- a/src/CrossValidation/Reports/ResidualAnalysis.php
+++ b/src/CrossValidation/Reports/ResidualAnalysis.php
@@ -49,7 +49,7 @@ class ResidualAnalysis implements Report
 
         $muHat = Stats::mean($labels);
 
-        $errors = $l1 = $l2 = [];
+        $errors = $l1 = $l2 = $log = [];
 
         $sse = $sst = 0.;
 
@@ -60,6 +60,7 @@ class ResidualAnalysis implements Report
 
             $l1[] = abs($error);
             $l2[] = $t = $error ** 2;
+            $log[] = log((1 + $label) / (1 + $prediction)) ** 2;
 
             $sse += $t;
             $sst += ($label - $muHat) ** 2;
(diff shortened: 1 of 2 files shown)


src/CrossValidation/Reports/ResidualAnalysis.php
```diff
@@ -60,6 +60,7 @@ public function generate(array $predictions, array $labels) : array
 
             $l1[] = abs($error);
             $l2[] = $t = $error ** 2;
+            $log[] = (log(1 + $label) - log(1 + $prediction)) ** 2;
```
andrewdalpino: Can this be optimized?

`(log((1 + $label) / (1 + $prediction))) ** 2`

andrewdalpino (changes requested):
Thanks for the PR, I left one comment regarding optimizing the SLE computation

Other than that, it looks good

src/CrossValidation/Reports/ResidualAnalysis.php
akondas: At the moment I don't know another solution.
msle_01

andrewdalpino: Since log(x) - log(y) is the same as log(x / y) ... can we arrange the formula such that we only have to compute the log once for each MSLE?

Ex.
```
(log((1 + $prediction) / (1 + $label))) ** 2
```

akondas: Yep, we can :+1:

akondas:
Lol, thanks anyway. A very nice example for others :clap:

andrewdalpino:
Your PR was denied because you violated this community's anti plagiarism policy by trying to claim code that I wrote as your own. I'll refer to the issue I posted (which you continually have tried to cover up) [here](https://github.com/php-ai/php-ml/issues/379)

Quoting our plagiarism policy ...

> Our community has a strong stance against plagiarism, or the copying of another author's code without attribution. Since the spirit of open source is to make code freely available, it is up to the community to enforce policies that deter plagiarism. As such, we do not allow contributions from those who violate this policy.

https://github.com/RubixML/RubixML/blob/master/CONTRIBUTING.md

PR #63, opened 2020-03-29
Fix `$estimate->proba` not preprocessing dataset in pipeline

Jord-JD:
When calling the `proba` method on the Pipeline estimator, the passed dataset is not preprocessed (transformed).

This PR fixes this by adding the preprocess call, as in the `predict` method.


andrewdalpino (approved):
Nice catch @DivineOmega 

Thank you for contributing!

We'll need to do the same for the `rank()` method as well

andrewdalpino:
I took care of the `rank()` method in  the last commit

Thank you again, this fix was pretty clutch

Feel free to join our chat group on Telegram at https://t.me/RubixML

PR #69, opened 2020-05-26
Perform a type check upon deserialization

ChristophWurst:
Otherwise any garbage data will cause a type error due to the mismatching return value and the return type hint.

### DIFF
--- a/src/Persisters/Serializers/Native.php
+++ b/src/Persisters/Serializers/Native.php
@@ -3,6 +3,10 @@
 namespace Rubix\ML\Persisters\Serializers;
 
 use Rubix\ML\Persistable;
+use RuntimeException;
+
+use function get_class;
+use function is_object;
 
 /**
  * Native
@@ -30,10 +34,18 @@ class Native implements Serializer
      * Unserialize a persistable object and return it.
      *
      * @param string $data
+     * @throws RuntimeException
      * @return \Rubix\ML\Persistable
      */
     public function unserialize(string $data) : Persistable
     {
-        return unserialize($data);
+        $unserialized = unserialize($data);
+        if (!is_object($unserialized)) {
+            throw new RuntimeException('Unserialized data is not an object');
+        }
+        if (!($unserialized instanceof Persistable)) {
+            throw new RuntimeException('Unserialized object is not a ' . Persistable::class . '. Got ' . get_class($unserialized));
+        }
+        return $unserialized;
     }
 }
--- a/tests/Persisters/Serializers/NativeTest.php
+++ b/tests/Persisters/Serializers/NativeTest.php
@@ -7,6 +7,10 @@ use Rubix\ML\Classifiers\DummyClassifier;
 use Rubix\ML\Persisters\Serializers\Native;
 use Rubix\ML\Persisters\Serializers\Serializer;
 use PHPUnit\Framework\TestCase;
+use RuntimeException;
+use stdClass;
+
+use function serialize;
 
 /**
  * @group Serializers
@@ -23,7 +27,7 @@ class NativeTest extends TestCase
      * @var \Rubix\ML\Persisters\Serializers\Native
      */
     protected $serializer;
-    
+
     /**
      * @before
      */
@@ -33,7 +37,7 @@ class NativeTest extends TestCase
 
         $this->persistable = new DummyClassifier();
     }
-    
+
     /**
      * @test
      */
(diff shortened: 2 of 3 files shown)


src/Persisters/Serializers/Native.php
```diff
@@ -3,6 +3,9 @@
 namespace Rubix\ML\Persisters\Serializers;
 
 use Rubix\ML\Persistable;
+use RuntimeException;
+use function get_class;
+use function is_object;
```
andrewdalpino, line 9: Classy function imports, nice touch, we generally put a space between the last class import and the first function import (same with imported constants). Check the Travis logs for the messages coming from cs-fixer.

`$ composer fix` will automatically format the code

https://travis-ci.org/github/RubixML/RubixML/builds/691331649

tests/Persisters/Serializers/NativeTest.php
```diff
@@ -42,19 +45,41 @@ public function build() : void
...
+        $this->expectException(RuntimeException::class);
+
+        $this->serializer->unserialize($data);
+    }
+
+    /**
+     * @test
+     */
+    public function deserializeNonPersistable() : void
+    {
```
andrewdalpino, line 84: Is this 'essentially' a duplicate test as the one above? Could it be factored into one test or perhaps a data provider would be helpful (in which case the method name could be something like deserializeBadData?

tests/Persisters/Serializers/NativeTest.php
```diff
@@ -7,6 +7,9 @@
 use Rubix\ML\Persisters\Serializers\Native;
 use Rubix\ML\Persisters\Serializers\Serializer;
 use PHPUnit\Framework\TestCase;
+use RuntimeException;
+use stdClass;
```
andrewdalpino, line 11: Space here

Run `$ composer check` to run the style checker, or `$composer fix` to automatically format the code per our style guide.

andrewdalpino (changes requested):
Nice catch @ChristophWurst

Had a question about the tests and Travis had a few things to say as well but other than that it looks great. Thanks for the PR welcome to Rubix

tests/Persisters/Serializers/NativeTest.php
ChristophWurst: I copy-pasted and forgot to change `stdClass` to something that is not an object. In any case, combining the two similar methods makes sense.

tests/Persisters/Serializers/NativeTest.php
```diff
@@ -42,19 +46,45 @@ public function build() : void
...
 
         $persistable = $this->serializer->unserialize($data);
 
         $this->assertInstanceOf(DummyClassifier::class, $persistable);
         $this->assertInstanceOf(Persistable::class, $persistable);
     }
+
+	/**
+	 * @return array<mixed>
+	 */
```
andrewdalpino: lol you made the change before i could comment ... this works, but should the annotation technically be `array[]`?

btw `$ composer analyze` to run static analysis

tests/Persisters/Serializers/NativeTest.php
```diff
@@ -42,19 +46,45 @@ public function build() : void
...
         $this->assertIsString($data);
 
         $persistable = $this->serializer->unserialize($data);
 
         $this->assertInstanceOf(DummyClassifier::class, $persistable);
         $this->assertInstanceOf(Persistable::class, $persistable);
     }
+
+    /**
+     * @return array<mixed>
```
andrewdalpino, line 66: lol made changes before I could drop the last comment ... this works, `array[]` perhaps could be more specific ... not a big deal

andrewdalpino (approved):

tests/Persisters/Serializers/NativeTest.php
ChristophWurst: phpstan inspired me to this :shrug: :speak_no_evil:

andrewdalpino:
Thanks again @ChristophWurst 

Feel free to join our chat room on Telegram https://t.me/RubixML

ChristophWurst:
Cheers!

I don't have Telegram, sorry :)

tests/Persisters/Serializers/NativeTest.php
andrewdalpino: I love Stan. Check our contributing guide if you haven't already ... we've got all the custom composer commands listed, we try to use static analysis as much as possible.

https://github.com/RubixML/RubixML/blob/master/CONTRIBUTING.md

PR #70, opened 2020-05-26
Handle multibyte string in text normalizer

maximecolin:
## Problem

`TextNormalizer` is unable to deal with string containing accents, chars like "é" or "è" are replaced by `?`.

Input: "Depuis qu’il avait emménagé à côté de chez elle, il y a de ça cinq ans."
Output: "depuis qu’il avait emm?nag? ? c?t? de chez elle, il y a de ?a cinq ans."

## Fix

Use `mb_strtolower` instead of `strtolower`.

### DIFF
--- a/benchmarks/Transformers/MultibyteTextNormalizerBench.php
+++ b/benchmarks/Transformers/MultibyteTextNormalizerBench.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Rubix\ML\Benchmarks\Transformers;
+
+use Rubix\ML\Datasets\Unlabeled;
+use Rubix\ML\Transformers\MultibyteTextNormalizer;
+
+/**
+ * @Groups({"Transformers"})
+ * @BeforeMethods({"setUp"})
+ */
+class MultibyteTextNormalizerBench
+{
+    protected const DATASET_SIZE = 100000;
+
+    /**
+     * @var \Rubix\ML\Datasets\Unlabeled
+     */
+    public $dataset;
+
+    /**
+     * @var \Rubix\ML\Transformers\TextNormalizer
+     */
+    protected $transformer;
+
+    public function setUp() : void
+    {
+        $text = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec at nisl posuere, luctus sapien vel, maximus ex. Curabitur tincidunt, libero at commodo tempor, magna neque malesuada diam, vel blandit metus velit quis magna. Vestibulum auctor libero quam, eu ullamcorper nulla dapibus a. Mauris id ultricies sapien. Integer consequat mi eget vehicula vulputate. Mauris cursus nisi non semper dictum. Quisque luctus ex in tortor laoreet tincidunt. Vestibulum imperdiet purus sit amet sapien dignissim elementum. Mauris tincidunt eget ex eu laoreet. Etiam efficitur quam at purus sagittis hendrerit. Mauris tempus, sem in pulvinar imperdiet, lectus ipsum molestie ante, id semper nunc est sit amet sem. Nulla at justo eleifend, gravida neque eu, consequat arcu. Vivamus bibendum eleifend metus, id elementum orci aliquet ac. Praesent pellentesque nisi vitae tincidunt eleifend. Pellentesque quis ex et lorem laoreet hendrerit ut ac lorem. Aliquam non sagittis est.';
+        $words = array_merge(['    ', '  ', '      '], explode(' ', $text));
+        $samples = [];
+
+        for ($i = 0; $i < self::DATASET_SIZE; $i++) {
+            shuffle($words);
+            $samples[] = [implode(' ', $words)];
+        }
+
+        $this->dataset = new Unlabeled($samples);
+
+        $this->transformer = new MultibyteTextNormalizer();
+    }
+
+    /**
+     * @Subject
+     * @Iterations(3)
+     * @OutputTimeUnit("seconds", precision=3)
+     */
+    public function apply() : void
+    {
+        $this->dataset->apply($this->transformer);
+    }
+}
--- a/benchmarks/Transformers/MultibyteTextNormalizerBench.php
+++ b/benchmarks/Transformers/MultibyteTextNormalizerBench.php
@@ -19,7 +19,7 @@ class MultibyteTextNormalizerBench
     public $dataset;
 
     /**
-     * @var \Rubix\ML\Transformers\TextNormalizer
+     * @var \Rubix\ML\Transformers\MultibyteTextNormalizer
      */
     protected $transformer;
 
--- a/benchmarks/Transformers/TextNormalizerBench.php
+++ b/benchmarks/Transformers/TextNormalizerBench.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Rubix\ML\Benchmarks\Transformers;
+
+use Rubix\ML\Datasets\Unlabeled;
+use Rubix\ML\Transformers\TextNormalizer;
+
+/**
+ * @Groups({"Transformers"})
+ * @BeforeMethods({"setUp"})
+ */
+class TextNormalizerBench
+{
+    protected const DATASET_SIZE = 100000;
+
+    /**
+     * @var \Rubix\ML\Datasets\Unlabeled
+     */
+    public $dataset;
+
+    /**
+     * @var \Rubix\ML\Transformers\TextNormalizer
+     */
+    protected $transformer;
+
+    public function setUp() : void
+    {
+        $text = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec at nisl posuere, luctus sapien vel, maximus ex. Curabitur tincidunt, libero at commodo tempor, magna neque malesuada diam, vel blandit metus velit quis magna. Vestibulum auctor libero quam, eu ullamcorper nulla dapibus a. Mauris id ultricies sapien. Integer consequat mi eget vehicula vulputate. Mauris cursus nisi non semper dictum. Quisque luctus ex in tortor laoreet tincidunt. Vestibulum imperdiet purus sit amet sapien dignissim elementum. Mauris tincidunt eget ex eu laoreet. Etiam efficitur quam at purus sagittis hendrerit. Mauris tempus, sem in pulvinar imperdiet, lectus ipsum molestie ante, id semper nunc est sit amet sem. Nulla at justo eleifend, gravida neque eu, consequat arcu. Vivamus bibendum eleifend metus, id elementum orci aliquet ac. Praesent pellentesque nisi vitae tincidunt eleifend. Pellentesque quis ex et lorem laoreet hendrerit ut ac lorem. Aliquam non sagittis est.';
+        $words = array_merge(['    ', '  ', '      '], explode(' ', $text));
+        $samples = [];
+
+        for ($i = 0; $i < self::DATASET_SIZE; $i++) {
+            shuffle($words);
+            $samples[] = [implode(' ', $words)];
+        }
+
+        $this->dataset = new Unlabeled($samples);
+
+        $this->transformer = new TextNormalizer();
+    }
+
+    /**
+     * @Subject
+     * @Iterations(3)
+     * @OutputTimeUnit("seconds", precision=3)
+     */
+    public function apply() : void
+    {
+        $this->dataset->apply($this->transformer);
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "phpunit/phpunit": "8.5.*"
     },
     "suggest": {
-        "ext-mbstring": "For MultibyteTextNormalizer",
+        "ext-mbstring": "For fast multibyte string manipulation.",
         "ext-tensor": "For fast Matrix/Vector computing",
         "ext-svm": "For Support Vector Machine engine (libsvm)",
         "ext-gd": "For image vectorization",
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,8 @@
         "php": ">=7.2",
         "amphp/parallel": "^1.3",
         "psr/log": "^1.1",
-        "rubix/tensor": "^2.0"
+        "rubix/tensor": "^2.0",
+        "symfony/polyfill-mbstring": "~1.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "2.16.*",
@@ -41,6 +42,7 @@
         "phpunit/phpunit": "8.5.*"
     },
     "suggest": {
+        "ext-mbstring": "For MultibyteTextNormalizer",
         "ext-tensor": "For fast Matrix/Vector computing",
         "ext-svm": "For Support Vector Machine engine (libsvm)",
         "ext-gd": "For image vectorization",
--- a/docs/transformers/multibyte-text-normalizer.md
+++ b/docs/transformers/multibyte-text-normalizer.md
@@ -1,0 +1,25 @@
+<span style="float:right;"><a href="https://github.com/RubixML/RubixML/blob/master/src/Transformers/MultibyteTextNormalizer.php">[source]</a></span>
+
+# Multibyte Text Normalizer
+This transformer converts all [multibyte text](https://www.php.net/manual/en/intro.mbstring.php) to lowercase and removes extra whitespace.
+
+Multibyte string contains multibyte characters sush as accented characters (é, è, à, ...), emojis (😀, 😉, ...) or characters of non roman alphabets (Chinese, Cyrillic, ...). 
+
+**Interfaces:** [Transformer](api.md#transformer)
+
+**Data Type Compatibility:** Categorical
+
+> **Note:** ⚠️ We recommend you install the [Multibyte string extension](https://www.php.net/manual/en/book.mbstring.php) for best performance.
+
+## Parameters
+This transformer does not have any parameters.
+
+## Additional Methods
+This transformer does not have any additional methods.
+
+## Example
+```php
+use Rubix\ML\Transformers\MultibyteTextNormalizer;
+
+$transformer = new MultibyteTextNormalizer();
+```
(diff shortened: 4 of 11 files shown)


ChristophWurst (changes requested):
Makes sense, though we'd need `ext-mbstring` at https://github.com/RubixML/RubixML/blob/5300b98a98c0041383cf747dd13cf5dff56c134d/composer.json#L30-L35 because that would be a hard dependency then. The alternative is to check for the module and only use `mb_*` when it's loaded.

mfn:
Not sure if it makes sense in this context here (performance?) but there would also be a polyfill available => https://github.com/symfony/polyfill-mbstring

andrewdalpino:
@maximecolin thanks for the PR

Are you aware of a performance difference between the two functions? We have a benchmarking suite perhaps that may be of service in this case. I would also be interested to know how the polyfill compares as well.

https://github.com/RubixML/RubixML/tree/master/benchmarks

`$ composer bench` to run the benchmarks or `$ ./vendor/bin/phpbench run ./benchmarks/Transformers/TextNormalizer.php` for example

Thanks again 2 in a row!

maximecolin:
Hi ! I created and ran a benchmark and it seems there is no noticeable performance difference between `strtolower` and `mb_strtolower`. Tested with 100000 samples, 10 iterations, on PHP 7.4.

With `strtolower`: 

```
#0  0.006 0.005 0.005 0.005 0.005 0.005 0.005 0.005 0.005 0.005 (s) [μ Mo]/r: 0.005 0.005 μRSD/r: 5.12%

1 subjects, 10 iterations, 1 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 0.005 [0.005 0.005] 0.006 (s)
⅀T: 0.050s μSD/r 0.000s μRSD/r: 5.121%
```

With `mb_strtolower`:

```
#0  0.006 0.005 0.005 0.005 0.005 0.005 0.005 0.005 0.005 0.005 (s) [μ Mo]/r: 0.005 0.005 μRSD/r: 5.22%

1 subjects, 10 iterations, 1 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 0.005 [0.005 0.005] 0.006 (s)
⅀T: 0.049s μSD/r 0.000s μRSD/r: 5.215%
```

With the polyfill:

```
#0  0.005 0.005 0.005 0.005 0.005 0.005 0.005 0.005 0.005 0.005 (s) [μ Mo]/r: 0.005 0.005 μRSD/r: 3.76%

1 subjects, 10 iterations, 1 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 0.005 [0.005 0.005] 0.005 (s)
⅀T: 0.050s μSD/r 0.000s μRSD/r: 3.758%
```

However i'm not sure i'm doing it properly. [Can you look at it ?](https://github.com/RubixML/RubixML/pull/70/files#diff-397ae520deba00dd2d8dc2b57e04158e)

andrewdalpino:
Hey @maximecolin you had a tiny little mistake on this line ... wrapping the samples in an array is not needed here.

mistake

Having that said, I got some interesting results when I ran the benchmark on Windows compared to Linux. It seems mbstring is faster on windows (PHP 7.4.2) but slower on Linux (WSL1 Ubuntu PHP 7.2.24). The opposite is true for regular `strtolower()`. The fastest was regular `strtolower()` on Linux at 0.6 seconds per 100,000 samples.

Windows mbstring

windows_mbstring

Windows regular

windows_regular

Linux mbstring

linux_mbstring


Linux regular

linux_regular

Could you try the benchmarks again to see if you get similar results?

I want to get a better idea if we should offer the user the option between regular and mbstring, and if so should the options be a constructor dependency or should we create a MultiByteTextNormalizer? I'd like to try to avoid a hard dependency on mbstring if possible as well.

benchmarks/Transformers/TextNormalizerBench.php
```diff
@@ -0,0 +1,53 @@
...
+     * @var \Rubix\ML\Transformers\TextNormalizer
+     */
+    protected $transformer;
+
+    public function setUp() : void
+    {
+        $text = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec at nisl posuere, luctus sapien vel, maximus ex. Curabitur tincidunt, libero at commodo tempor, magna neque malesuada diam, vel blandit metus velit quis magna. Vestibulum auctor libero quam, eu ullamcorper nulla dapibus a. Mauris id ultricies sapien. Integer consequat mi eget vehicula vulputate. Mauris cursus nisi non semper dictum. Quisque luctus ex in tortor laoreet tincidunt. Vestibulum imperdiet purus sit amet sapien dignissim elementum. Mauris tincidunt eget ex eu laoreet. Etiam efficitur quam at purus sagittis hendrerit. Mauris tempus, sem in pulvinar imperdiet, lectus ipsum molestie ante, id semper nunc est sit amet sem. Nulla at justo eleifend, gravida neque eu, consequat arcu. Vivamus bibendum eleifend metus, id elementum orci aliquet ac. Praesent pellentesque nisi vitae tincidunt eleifend. Pellentesque quis ex et lorem laoreet hendrerit ut ac lorem. Aliquam non sagittis est.';
+        $words = array_merge(['    ', '  ', '      '], explode(' ', $text));
```
andrewdalpino, line 29: Slick setup @maximecolin I like how you added the extra whitespace

benchmarks/Transformers/TextNormalizerBench.php
```diff
@@ -0,0 +1,53 @@
...
+        $samples = [];
+
+        for ($i = 0; $i < self::DATASET_SIZE; $i++) {
+            shuffle($words);
+            $samples[] = [implode(' ', $words)];
+        }
+
+        $this->dataset = new Unlabeled([$samples]);
+
```
andrewdalpino, line 38: Array wrap is not needed here

maximecolin:
@andrewdalpino Ok, I fixed the benchmark and indeed, `mb_strtolower` is much slower than `strtolower` and even worst for the polyfill (tested on PHP 7.4 on macOS).

With `mb_strtolower`:

```
#0  1.827 1.803 1.815 1.819 1.813 1.804 1.788 1.825 1.830 1.791 (s) [μ Mo]/r: 1.812 1.818 μRSD/r: 0.77%

1 subjects, 10 iterations, 1 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 1.788 [1.812 1.818] 1.830 (s)
⅀T: 18.115s μSD/r 0.014s μRSD/r: 0.770%
```

With `strtolower`:

```
#0  0.857 0.839 0.854 0.867 0.839 0.833 0.847 0.839 0.830 0.912 (s) [μ Mo]/r: 0.852 0.843 μRSD/r: 2.69%

1 subjects, 10 iterations, 1 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 0.830 [0.852 0.843] 0.912 (s)
⅀T: 8.518s μSD/r 0.023s μRSD/r: 2.688%
```

With the polyfill:

```
#0  8.565 8.715 8.601 8.543 8.432 8.535 8.463 8.536 8.489 8.427 (s) [μ Mo]/r: 8.531 8.522 μRSD/r: 0.96%

1 subjects, 10 iterations, 1 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 8.427 [8.531 8.522] 8.715 (s)
⅀T: 85.306s μSD/r 0.082s μRSD/r: 0.961%
```

So I propose to move `ext-mbstring` in the `"suggest"` section of composer.json and to add a `bool $multibyte = false` parameter the `TextNormalizer` construct. If true, it uses `mb_strtolower` (or the polyfill if `ext-mbstring` is missing) else it uses `strtolower`. By default it keeps the performance high and if needed, users can enable multibyte. 

```php
new TextNormalizer(); // use strtolower
new TextNormalizer(true); // use mb_strtolower
```

What do you think?

maximecolin:
I just read your comment again and a `MultiByteTextNormalizer` could be fine too.

andrewdalpino:
What if we thought about it for a day or so @maximecolin? I'm leaning toward the separate class but I'd like to hear what you think. I see pros and cons either way.

maximecolin:
Well, i'm agree with you. I thing adding options will always add complexity, one option today, more options tomorrow and the class could be a big "if" tangle. In other hand, two classes will duplicate a lot of code, but maybe we can factorize some part of it.

ChristophWurst:
Put differently: how much of a performance hit would the switch cause realistically, like in a real project that uses this package. Is this something that is called often? Just that the multibyte version is slower does not mean all of the text normalization is much slower if that is used only a few times.

andrewdalpino:
I think we're on the same page @maximecolin ... granted I am by no means an expert on the mbstring extension - I use it, and that's about it.

Here's my thought process ... multibyte strings seems to be different from regular strings in the technical sense. Reading the docs (https://www.php.net/manual/en/ref.mbstring.php) it looks like you can perform some functions on multibyte string that you cannot on regular strings (possibly vice versa). For example, we might want to add this function in the future (https://www.php.net/manual/en/function.mb-convert-encoding.php) which is only an mbstring function. Just using that as an example to help explain what I mean. If we stuffed everything into one class, this could get ugly because for every function we call on a string (right now it's just `strtolower` but I'm thinking of the future) we'd have to check whether it should be the regular or mbstring version. In addition, there is a notable performance difference between the two.

I'm perfectly fine with a little code duplication if in return we get separation of concerns. Modularity is one of our core design principals. Based on my thought process above, I'm considering operations on multibyte strings to be a separate concern. In addition, we'll need logic to check for the mbstring extension which will only be necessary for the MultiByteTextNormalizer.

For NLP tasks TextNormalizer will be used quite often @ChristophWurst. You want to be able to have a canonical symbol for every word in the corpus unless for some reason 'The' should be treated differently from 'the'. It looks like mbstring is about twice as slow and that is enough to make a difference in my opinion. That difference increases by an order of magnitude (factor of 10) if the user does not have the mbstring extension enabled. Having two classes makes it easier to deprecate one for the other if indeed this difference is found to be negligible in the future, however.

What do you guys think?

ChristophWurst:
Makes sense. Thanks for the details.

IMO for proper i18n support of languages other than English and also given the popularity of emoji in texts nowadays, the multibyte support is an absolute necessity as the traditional php string functions are unpredictable with those. I do, however, see that the performance does matter in some cases, so it would make sense to have both. Though the users of the lib should be warned that the non-mb version might give they inaccurate results. This is typically something you have to learn the hard way, like when your testing data does not contain any special characters but the real world data does.

maximecolin:
Totally agreed on the separation of concerns. Going for two classes.

> we'll need logic to check for the mbstring extension

@andrewdalpino so you thing should raise an exception if the mbstring extension is missing? Or a warning is enough as the polyfill will make the normalizer work without the extension?

andrewdalpino:
@maximecolin  Oh that's right I forgot about the polyfill ... we should be good to go then! Also we're justified in adding the polyfill as a project dependency as I completely agree with @ChristophWurst that it's important for proper i18n support. Emojis are THE FUTURE. Before you know it our entire language will be built up of nothing but emojis.

Btw I wanted to mention, we have an [Extras](https://github.com/RubixML/Extras) package for experimental features. This transformer could also go there if think it's more appropriate. I'm down for putting it either place as, even though we're under feature freeze until 0.1.0, it's a smaller feature.

ChristophWurst:
> Emojis are THE FUTURE. Before you know it our entire language will be built up of nothing but emojis.

Now I'm not sure if you're just joking or making fun of my statement. You have to use more emojis so I can get your intentions :stuck_out_tongue_winking_eye:

andrewdalpino:
Joking ... we'll always need letters to spell out proper names ... but everything else will be emojis though 🙂

BTW thanks alot guys this is the most action my repo has seen in a while lol

andrewdalpino:
@maximecolin Do you think you can handle the documentation or would you like me to take care of that? Here are the docs for TextNormalizer ... https://github.com/RubixML/RubixML/blob/master/docs/transformers/text-normalizer.md

I was thinking we could put a note about slower performance when the mbstring extension is not loaded. And any other caveats, etc.

maximecolin:
@andrewdalpino I totally can handle de documentation update

andrewdalpino:
Awesome!

Don't forget to update https://github.com/RubixML/RubixML/blob/master/mkdocs.yml and https://github.com/RubixML/RubixML/blob/master/docs/preprocessing.md (under Text Cleaning).

EDIT: Also I think we should recommend the mbstring extension in the README and docs index (I'd suggest next to the GD extension)

https://github.com/RubixML/RubixML/blob/master/README.md
https://github.com/RubixML/RubixML/blob/master/docs/index.md

EDIT 2: and also in the suggest section of composer.json

docs/transformers/text-normalizer.md
```diff
@@ -17,5 +19,5 @@ This transformer does not have any additional methods.
 ```php
 use Rubix\ML\Transformers\TextNormalizer;
 
-$transformer = new TextNormalizer(true);
+$transformer = new TextNormalizer();
```
andrewdalpino, line 22: Nice catch ... we used to allow the user to turn on/off removing extra whitespace

src/Transformers/MultibyteTextNormalizer.php
```diff
@@ -0,0 +1,57 @@
...
+use Rubix\ML\DataType;
+
+/**
+ * Multibyte Text Normalizer
+ *
+ * This transformer converts all multibyte text to lowercase and removes extra whitespace.
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Andrew DalPino
```
andrewdalpino: This should be your name :)

docs/transformers/text-normalizer.md
```diff
@@ -7,6 +7,8 @@ This transformer converts all text to lowercase and removes extra whitespace.
 
 **Data Type Compatibility:** Categorical
 
+⚠️ This transformer can't handle multibyte text properly. For multibyte text, use [MultibyteTextNormalizer](multibyte-text-normalizer.md).
```
andrewdalpino: I like this warning ... can we format it like this to stay consistent with the rest of the docs?

> **Note:** ⚠️ This transformer can't handle multibyte text properly. For multibyte text, use [MultibyteTextNormalizer](multibyte-text-normalizer.md).

Example

https://github.com/RubixML/RubixML/edit/master/docs/classifiers/adaboost.md

docs/transformers/multibyte-text-normalizer.md
```diff
@@ -0,0 +1,23 @@
+<span style="float:right;"><a href="https://github.com/RubixML/RubixML/blob/master/src/Transformers/MultibyteTextNormalizer.php">[source]</a></span>
+
+# Multibyte Text Normalizer
+This transformer converts all multibyte text to lowercase and removes extra whitespace.
+
+**Interfaces:** [Transformer](api.md#transformer)
+
+**Data Type Compatibility:** Categorical
+
+⚠ This normalizer is slower than the non multibyte [TextNormalizer](text-normalizer.md). Using the multibyte string polyfill will also strongly reduce performance of the transformer. We recommend to install the [multibyte string extension](https://www.php.net/manual/en/book.mbstring.php).
```
andrewdalpino: I think we can simplify this note to just "We recommend you install the Multibyte string extension for best performance."

docs/transformers/multibyte-text-normalizer.md
```diff
@@ -0,0 +1,23 @@
+<span style="float:right;"><a href="https://github.com/RubixML/RubixML/blob/master/src/Transformers/MultibyteTextNormalizer.php">[source]</a></span>
+
+# Multibyte Text Normalizer
+This transformer converts all multibyte text to lowercase and removes extra whitespace.
```
andrewdalpino: For those of us who have no idea what a multibyte string is, perhaps we can include a sentence about that?

composer.json
```diff
@@ -41,6 +42,7 @@
         "phpunit/phpunit": "8.5.*"
     },
     "suggest": {
+        "ext-mbstring": "For MultibyteTextNormalizer",
```
andrewdalpino: We may use multibyte strings in other parts of the library so this might not always be the case and, if so, I guarantee I'll forget to change this line to something else. Perhaps we can make a more general statement such as "For fast multibyte string manipulation."

Open to other catchy phrases

andrewdalpino (changes requested):
Very nice work @maximecolin, I left a couple comments about expanding the docs and some minor non-code related things but other than that it looks good to go!

Also Travis had a few things to say https://travis-ci.org/github/RubixML/RubixML/jobs/692483989

Just an FYI, all the composer commands to run static analysis, code style fixer, unit tests, etc can be found here https://github.com/RubixML/RubixML/blob/master/CONTRIBUTING.md

`$ composer build` to run everything (this is also what Travis runs)

benchmarks/Transformers/MultibyteTextNormalizerBench.php
```diff
@@ -0,0 +1,51 @@
...
+    protected const DATASET_SIZE = 100000;
+
+    /**
+     * @var \Rubix\ML\Datasets\Unlabeled
+     */
+    public $dataset;
+
+    /**
+     * @var \Rubix\ML\Transformers\TextNormalizer
+     */
```
andrewdalpino, line 23: This should be `@var \Rubix\ML\Transformers\MultibyteTextNormalizer`

tests/Transformers/MultibyteTextNormalizerTest.php
```diff
@@ -0,0 +1,68 @@
...
+class MultibyteTextNormalizerTest extends TestCase
+{
+    /**
+     * @var \Rubix\ML\Datasets\Unlabeled
+     */
+    protected $dataset;
+
+    /**
+     * @var \Rubix\ML\Transformers\TextNormalizer
+     */
```
andrewdalpino, line 23: This should be @var \Rubix\ML\Transformers\MultibyteTextNormalizer

This *should* also fix the second error that Stan found on line 47

andrewdalpino (approved):

andrewdalpino:
Thanks again @maximecolin!

Feel free to join out chat room on Telegram https://t.me/RubixML

PR #77, opened 2020-06-13
Better keywords for findability on Packagist.org

marclaporte:
1- Because of https://github.com/composer/packagist/issues/1091 we are better off to use spaces than hyphens. I left k-means and k-nearest because that is the proper spelling.
2- I added a few more relevant keyworks which are used on Packagist.org

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,10 +6,11 @@
     "license": "MIT",
     "readme": "README.md",
     "keywords": [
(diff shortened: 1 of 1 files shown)


andrewdalpino:
LGTM!

marclaporte:
Better handling upstream: https://github.com/composer/packagist/commit/1bb3e84ff368d89881fa75c435c413b9caf291ae

PR #85, opened 2020-07-08
Add RegexFilter::EXTRA_WORDS

### LINKED ISSUE #84: 'Extra words' pattern to match repeated words for Regex Filter
The [RegexFilter](https://docs.rubixml.com/en/latest/transformers/regex-filter.html) is a handy [Transformer](https://docs.rubixml.com/en/latest/transformers/api.html#transformer) that removes bits of text from a corpus that match a set of user-defined regular expression patterns. The class provides a number of pre-defined regex patterns as class constants including URL, email address, and extra character matching. This task is to implement a case-insensitive 'extra words' pattern to remove repeated words in a string.

**Example**

```
Yo yo, what is up my my man?
```

should filter to something like ...

```
Yo , what is up my  man?
```

simplechris:
###### SUMMARY

Hey 👋 . This small PR adds `RegexFilter::EXTRA_WORDS`, to filter out consecutively repeated words.

----

###### LINKS

- Closes #84

### DIFF
--- a/tests/Transformers/RegexFilterTest.php
+++ b/tests/Transformers/RegexFilterTest.php
@@ -32,6 +32,8 @@ class RegexFilterTest extends TestCase
             ['I was not proud of what I had learned, but I never doubted that it was worth $$$ knowing..'],
             ['Too weird to live, support@rubixml.com too rare to die https://rubixml.com'],
             ['A man who procrastinates in @his choosing will inevitably have his choice made for him by #circumstance'],
+            ['The quick quick brown fox jumped over the lazy man sitting at a bus stop drinking a can of Cola cola'],
+            ['Diese äpfel Äpfel schmecken sehr gut'],
         ]);
 
         $this->transformer = new RegexFilter([
@@ -40,6 +42,7 @@ class RegexFilterTest extends TestCase
             RegexFilter::MENTION,
             RegexFilter::HASHTAG,
             RegexFilter::EXTRA_CHARACTERS,
+            RegexFilter::EXTRA_WORDS,
         ]);
     }
 
@@ -63,6 +66,8 @@ class RegexFilterTest extends TestCase
             ['I was not proud of what I had learned, but I never doubted that it was worth $ knowing.'],
             ['Too weird to live,  too rare to die '],
             ['A man who procrastinates in  choosing will inevitably have his choice made for him by '],
+            ['The  quick brown fox jumped over the lazy man sitting at a bus stop drinking a can of  cola'],
+            ['Diese  Äpfel schmecken sehr gut'],
         ];
 
         $this->assertEquals($expected, $this->dataset->samples());
--- a/docs/transformers/regex-filter.md
+++ b/docs/transformers/regex-filter.md
@@ -34,6 +34,7 @@ $transformer = new RegexFilter([
 | MENTION | A pattern that matches Twitter-style mentions (@example). |
 | HASHTAG | Matches Twitter-style hashtags (#example). |
 | EXTRA_CHARACTERS | Matches extra non word or number characters such as repeated punctuation and special characters. |
+| EXTRA_WORDS | Matches extra (consecutively repeated) words.
 
 ## Additional Methods
 This transformer does not have any additional methods.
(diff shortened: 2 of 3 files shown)


andrewdalpino:
Hey @simplechris thanks for the PR! How are you doing?

What I had in mind was a class constant such as the one here https://github.com/RubixML/RubixML/blob/master/src/Transformers/RegexFilter.php#L75. The idea is the user could use the constant in the class constructor.

**Example**

```php
use Rubix\ML\Transformers\RegexFilter;

$transformer = new RegexFilter([
    RegexFilter::EXTRA_WORDS,
    // Other patterns ...
]);
```

Is there a way to convert the pattern you created in this class into something that would be compatible with RegexFilter as a class constant?

P.S. You can run `$ composer fix` to automatically fix all the coding style issues. Our [contributing guide](https://github.com/RubixML/RubixML/blob/master/CONTRIBUTING.md) has all the composer commands we use in the project.

simplechris:
Yeah, I had look at something like that initially, but the pattern I introduced needs to pass a second argument to the `preg_replace` call, I don't _think_ the current implementation of `RegexFilter` allows for this.

I'd need something like:

```php
preg_replace(self::EXTRA_WORDS, '$1', $value)
```

Not impossible, I'll take another look after work later to see if I can come up with something that doesn't introduce breaking changes. Maybe also (optionally) accepting input like this:

```php
$transformer = new RegexFilter([
    [RegexFilter::EXTRA_WORDS, '$1'],
    // Other patterns ...
]);
```

Equally, I may have just looked at this with tired eyes last night and be missing the obvious 🙈

andrewdalpino:
Sounds good, I'm not a regex expert but I assumed that there would be a way to select/match all repeat words after the first word thus filtering them by replacing them with a `''`.

Also, feel free to join us in our Telegram channel https://t.me/RubixML

simplechris:
@andrewdalpino I've added an implementation that just uses `RegexFilter`. I agree that it's nicer than a separate `ExtraWordsFilter`, so i'll go ahead and remove that now... 

>  I assumed that there would be a way to select/match all repeat words after the first word thus filtering them by replacing them with a `''`.

Couldn't find a way to get it working specially like this, but check out the implementation in 677d7c0. It simply looks in the replacementMap for a value to use, or defaults back to `''` as before.

andrewdalpino:
Hey @simplechris! Do you know of a regex pattern that would allow us to use the current replacement string (`''`)? For example, is there a pattern that can just match the words after the first word? From my understanding, if we can do this, we can use the current strategy of just replacing the matched words with `''`.

The reason I think we should give it another look (again, I'm not strong in regex for the same reasons as most) is because I was able to do this with the EXTRA_CHARACTERS pattern (https://github.com/RubixML/RubixML/blob/master/src/Transformers/RegexFilter.php#L75) and my reasoning is that if we can do it for characters we can also do it for words. Is that a cogent assessment?

simplechris:
Yeah, i'm no regex master either, but I've now got something working using a positive lookahead in a similar manner to the existing `EXTRA_CHARACTERS`. I've rebased this PR to remove the previous implementation 👍

tests/Transformers/RegexFilterTest.php
```diff
@@ -32,6 +32,7 @@ protected function setUp() : void
             ['I was not proud of what I had learned, but I never doubted that it was worth $$$ knowing..'],
             ['Too weird to live, support@rubixml.com too rare to die https://rubixml.com'],
             ['A man who procrastinates in @his choosing will inevitably have his choice made for him by #circumstance'],
+            ['The quick quick brown fox jumped over the lazy man sitting at a bus stop.'],
```
andrewdalpino: Can we also test repeated words in different case? Ex. "Yo yo what's up?"

andrewdalpino (changes requested):
Looks great Chris! I'm happy we found a solution that works with the current setup. I had one request to test words of different cases but other than that LGTM.

tests/Transformers/RegexFilterTest.php
simplechris: Ahh yeah, sure. As the regex includes the case insensitive (`i`) unicode (`u`) flags i'll ensure test cases for both these variants too.

Issue #99, opened 2020-08-08
PHP Fatal error:  Uncaught InvalidArgumentException: MLP Regressor when loading labeled dataset from iterator 


ghost:
**Describe the bug**

Hi there, I just prepared a basic [`dataset/1_1_beginner.csv`](https://github.com/programarivm/pgn-chess-data/blob/master/dataset/1_1_beginner.csv) file in order to train a beginner chess model in batches:

```
php cli/prepare/beginner.php 1 1
```
This CSV file contains the samples and labels of the first game in the database in order to keep things simple.

```
0;34;2;1;40.06;0;563.66
0;36;2;1;40.06;0;574.66
0;36;2;1;40.06;0;574.66
0;33;2;1;40.06;0;565.66
0;32;2;1;40.06;0;568.66
0;31;2;1;40.06;0;570.66
1;30;2;1;40.06;0;569.66
1;29;3;1;40.06;0;542.66
1;31;3;1;40.06;0;543.66
1;29;2;1;39.06;0;567.66
1;31;3;1;39.06;0;563.66
1;31;3;1;39.06;0;563.66
1;32;3;1;39.06;0;566.66
2;28;5.2;1;39.06;0;570.66
...
```

**To Reproduce**

However, a PHP fatal error is thrown when using this CSV to run the [`cli/train/beginner.php`](https://github.com/programarivm/pgn-chess-data/blob/master/cli/train/beginner.php) script:

```php
<?php

require_once __DIR__ . '/../../vendor/autoload.php';

use Rubix\ML\PersistentModel;
use Rubix\ML\CrossValidation\Metrics\RSquared;
use Rubix\ML\Datasets\Labeled;
use Rubix\ML\Extractors\CSV;
use Rubix\ML\Extractors\ColumnPicker;
use Rubix\ML\NeuralNet\CostFunctions\LeastSquares;
use Rubix\ML\NeuralNet\Layers\Dense;
use Rubix\ML\NeuralNet\Layers\Activation;
use Rubix\ML\NeuralNet\ActivationFunctions\ReLU;
use Rubix\ML\NeuralNet\Optimizers\RMSProp;
use Rubix\ML\Other\Loggers\Screen;
use Rubix\ML\Persisters\Filesystem;
use Rubix\ML\Regressors\MLPRegressor;

const DATASET_FOLDER = __DIR__.'/../../dataset';
const MODEL_FOLDER = __DIR__.'/../../model';

[...]

PR #100, opened 2020-08-09
CSV example added to Extracting Data

ghost:
As discussed on https://github.com/RubixML/RubixML/issues/99

### DIFF
--- a/docs/extracting-data.md
+++ b/docs/extracting-data.md
@@ -38,7 +38,7 @@ $dataset = Labeled::fromIterator($extractor)
     ->transformLabels('floatval');
 ```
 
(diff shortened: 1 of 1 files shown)


andrewdalpino (changes requested):
Since this information primarily deals with preprocessing and since we already have a similar example in the docs, could we rather add a couple words or if needed a whole sentence to the 'preprocessing' section?

andrewdalpino:
Hey @programarivm I made some changes to the docs in this commit https://github.com/RubixML/RubixML/commit/d13cd3c61d90159b07bbf8b4ea5fdafc955bf352

It looks like we already had a similar example to yours and notes in the documentation about the labels so this should suffice

Thanks for the pull request!

PR #103, opened 2020-08-16
Fix readme doc links to those with CSS styling

michaelbutler:
Hi, simply updated three URLs from the README.md as I noticed the current links (those with `/en/latest` in the URL) were readable, but had no styling whatsoever (the CSS 404'd). I updated the links based on what I saw on https://docs.rubixml.com/ directly, and they had all the proper styling.

It looks like github doesn't show the inline diff on the first change (the badge icon) but all I changed there was the /en/latest as well.

# Testing

Simply click the links in question and confirm it goes to a page with CSS styling applied and the left sidebar is shown, etc.


andrewdalpino:
Nice catch @michaelbutler and thanks for the fix!

Drop in on our Telegram channel if you'd like https://t.me/RubixML

PR #105, opened 2020-08-26
Optimize Cosine distance kernel - skip normalizing vectors if product is zero

kroky:
A simple optimization for sparse vectors - we don't need to calculate the normalized vector forms if vector product is zero. Let me know if the zero comparison is OK, not sure I found how you do it in other places...

### DIFF
--- a/src/Kernels/Distance/Cosine.php
+++ b/src/Kernels/Distance/Cosine.php
@@ -52,9 +52,15 @@ class Cosine implements Distance, Stringable
             $valueB = $b[$i];
 
             $sigma += $valueA * $valueB;
+        }
 
-            $ssA += $valueA ** 2;
-            $ssB += $valueB ** 2;
+        if (abs($sigma) > EPSILON) {
+            foreach ($a as $i => $valueA) {
+                $valueB = $b[$i];
+
+                $ssA += $valueA ** 2;
+                $ssB += $valueB ** 2;
+            }
         }
 
         return 1.0 - ($sigma / (sqrt($ssA * $ssB) ?: EPSILON));


andrewdalpino (commented):
Hey @kroky I'm thinking what if we made a SparseCosine kernel? Are there other optimizations we can make knowing that the vectors are sparse? Will this give us the best of both worlds?

andrewdalpino:
Also, I think we'll be ok with just checking for 0 instead of +/- EPSILON since even after TF-IDF words that do not appear in a document are still 0.

Annotation 2020-08-26 194345

src/Kernels/Distance/Cosine.php
```diff
@@ -52,9 +52,15 @@ public function compute(array $a, array $b) : float
             $valueB = $b[$i];
 
             $sigma += $valueA * $valueB;
+        }
+
+        if (abs($sigma) > EPSILON) {
```
andrewdalpino, line 57: I think we'll be ok with just checking for 0 instead of +/- EPSILON since even after TF-IDF words that do not appear in a document are still 0.

Annotation 2020-08-26 194345

kroky:
I couldn't find more optimizations we can make besides making use of state and caching computations. SparseCosine is definitely possible but isn't it duplicating kernel distances? I mean, we will have 2 distances doing the same, one is just optimized for sparse vectors. Could confuse end users. I feel like Cosine itself should work well with both non-sparse and sparse vectors. Anyway, let me know if you insist and I can move this optimization out in a separate kernel class.

andrewdalpino:
I'm getting about a 2.2ms gain in wall time for dense vectors vs the previous version but a whopping 7.7ms loss in wall time for highly-sparse vectors (90% zeros). For kinda sparse (50%) I get a drop of 3.3ms off the wall time and for not really sparse (10%) there's no difference.

I also wonder if there are optimization that we can make in the future that only apply to sparse vectors. The 2.2ms plus the possibility of future optimizations guide me to suggest a new class. For now, I don't think it's a big deal but if we discover a better optimization in the future (that hurts dense vector performance even more than 2.2ms) we should create a separate class at that time.

Btw I experimented with some other optimizations as well. The ones that came close to beating your version were slightly slower for sparse vectors (4.7ms vs 4.0ms) and slightly faster for dense vectors (12.5ms vs. 14.7ms). Ex.

```php
    public function compute(array $a, array $b) : float
    {
        $sigma = $ssA = $ssB = 0.0;

        foreach ($a as $i => $valueA) {
            $valueB = $b[$i];

            if ($valueA != 0 and $valueB != 0) {
                $sigma += $valueA * $valueB;

                $ssA += $valueA ** 2;
                $ssB += $valueB ** 2;
            } else {
                if ($valueA != 0) {
                    $ssA += $valueA ** 2;
                }

                if ($valueB != 0) {
                    $ssB += $valueB ** 2;
                }
            }
        }

        return 1.0 - ($sigma / (sqrt($ssA * $ssB) ?: EPSILON));
    }
```

I like your version better though

Very nice work @kroky and thank you for the contribution!

andrewdalpino:
Bechmarks

https://github.com/RubixML/RubixML/blob/master/benchmarks/Kernels/Distance/CosineBench.php

cosine-sparsity-performance

Annotation 2020-08-28 010745

kroky:
Great, thanks! Nice benchmark comparison. I'd definitely consider caching computed results down the road to further optimize overall performance.

PR #107, opened 2020-09-03
Flysystem Persister

### LINKED ISSUE #104: Remote Filesystem (S3, Azure etc)
The Filesystem class states:
```
Filesystems are local or remote storage drives that are organized by files and folders. The filesystem persister serializes models to a file at a user-specified path.
```
I can't find anything in the documentation about how to use a remote model, for instance: read a model remotely and train it partially OR read a model remotely and perform predictions from it.

I'm considering a project on a serverless platform and won't be able to just download the model like I am currently. Anyone point me in the right direction? Thanks!

simplechris:
###### SUMMARY

This PR adds a [Flysystem](https://flysystem.thephpleague.com) [`Persistor`](https://docs.rubixml.com/en/latest/persisters/api.html). This enables a user to load and save models located in a remote storage backend (such as Amazon S3, Azure Blob Storage, Google Cloud Storage, Dropbox etc...)

Closes #104 .

---

> A proof-of-concept PR (#106) instigated some interesting discussion about how to extend the concept of persistence in RubixML. A wider discussion around the introduction of a flexible persistence-subsystem is now ongoing. In the meantime this Flysystem Persitor will allow connectivity to a large array of remote storage solutions 🚀

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
     "require": {
         "php": ">=7.2",
         "amphp/parallel": "^1.3",
+        "league/flysystem": "^1.1",
         "psr/log": "^1.1",
         "rubix/tensor": "^2.0",
         "symfony/polyfill-mbstring": "^1.0",
@@ -42,6 +43,7 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "2.16.*",
+        "league/flysystem-memory": "^1.0",
         "phpbench/phpbench": "0.17.*",
         "phpstan/phpstan": "0.12.*",
         "phpstan/phpstan-phpunit": "0.12.*",
--- a/docs/persisters/flysystem.md
+++ b/docs/persisters/flysystem.md
@@ -1,0 +1,94 @@
+<span style="float:right;"><a href="https://github.com/RubixML/RubixML/blob/master/src/Persisters/Flysystem.php">[source]</a></span>
+
+# Flysystem
+[Flysystem](https://flysystem.thephpleague.com) is a filesystem abstraction library providing a unified interface for 
+[many different filesystems](https://github.com/thephpleague/flysystem#adapters). It enables access to remote storage backends such as Amazon S3, Azure Blob Storage, Google Cloud Storage, Dropbox...
+
+The Flysystem persister saves models to a file at a given path using the Flysystem library, and can automatically keep a history of past saved models.
+
+## Parameters
+| # | Param | Default | Type | Description |
+|---|---|---|---|---|
+| 1 | path | | `string` | The path to the model file on the filesystem. |
+| 2 | filesystem |  | `FilesystemInterface` | The flysystem object providing access to your storage backend |
+| 3 | serializer | `Native` | `Serializer` | The serializer used to convert to and from storage format. |
+
+## Examples
+
+### Local:
+ Using the Flysystem Persister to interact with data stored on your local filesystem:
+- Pass the Flysystem Persister a `Filesystem` object that uses the `Local` adapter instance:
+```php
+use League\Flysystem\Filesystem;
+use League\Flysystem\Adapter\Local;
+use Rubix\ML\Persisters\Flysystem;
+
+$storage = new Filesystem(new Local('/'));
+$persister = new Flysystem('/path/to/example.model', $storage);
+
+// Or, to save keystrokes you could also use the shortcut method:
+
+$persister = Flysystem::local('/path/to/example.model');
+```
+
+### FTP Server:
+Using the Flysystem Persister to interact with data stored on an FTP Server:
+- Pass the Flysystem Persister a `Filesystem` object that uses the `Ftp` adapter instance:
+
+```php
+use League\Flysystem\Filesystem;
+use League\Flysystem\Adapter\Ftp;
+use Rubix\ML\Persisters\Flysystem;
+
+$config = [
+    'host' => 'ftp.example.com',
+    'username' => 'username',
+    'password' => 'password',
+];
+$storage = new Filesystem(new Ftp($config));
+$persister = new Flysystem('/path/to/example.model', $storage);
+
+// Or, to save keystrokes you could also use the shortcut method:
+
+$persister = Flysystem::ftp('/path/to/example.model', $config);
+```
+
+### Amazon S3
+Using the Flysystem Persister to interact with data on Amazon S3:
+- install the Flysystem S3 adapter: `composer require league/flysystem-aws-s3-v3`
+- Pass the Flysystem Persister a `Filesystem` object that uses the `AwsS3Adapter` adapter instance:
+
+```php
+use Aws\S3\S3Client;
+use League\Flysystem\Filesystem;
+use League\Flysystem\AwsS3v3\AwsS3Adapter;
+use Rubix\ML\Persisters\Flysystem;
+
+$client = new S3Client([
+    'credentials' => [
+        'key'    => 'your-key',
+        'secret' => 'your-secret',
+    ],
+    'region' => 'your-region',
+    'version' => 'latest|version',
+]);
+
+$storage = new Filesystem(new AwsS3Adapter($client, 'your-bucket-name'));
+
+$persister = new Flysystem('/path/to/example.model', $storage);
+```
+
+
+## Additional Methods
+
+Shortcut to return a Flysystem Persister backed by the [Local](https://flysystem.thephpleague.com/v1/docs/adapter/local/) filesystem
+
+```php
+public static local(string $path, bool $history = false, ?Serializer $serializer = null) : self
+```
+
+Shortcut to return a Flysystem Persister backed by an [FTP Server](https://flysystem.thephpleague.com/v1/docs/adapter/ftp/).
+
+```php
+public static ftp(string $path, array $config, bool $history = false, ?Serializer $serializer = null) : self
+```
--- a/src/Persisters/Flysystem.php
+++ b/src/Persisters/Flysystem.php
@@ -1,0 +1,189 @@
+<?php
+
+namespace Rubix\ML\Persisters;
+
+use League;
+use League\Flysystem\Adapter\Local;
+use League\Flysystem\Adapter\Ftp;
+use League\Flysystem\FilesystemInterface;
+use Rubix\ML\Encoding;
+use Rubix\ML\Persistable;
+use Rubix\ML\Other\Helpers\Params;
+use Rubix\ML\Persisters\Serializers\Native;
+use Rubix\ML\Persisters\Serializers\Serializer;
+use RuntimeException;
+use Stringable;
+
+/**
+ * Flysystem
+ *
+ * The flysystem persister serializes models to a file at a user-specified path, using a
+ * user-provided flysystem instance.
+ *
+ * Flysystem is a filesystem abstraction providing a unified interface for many different
+ * filesystems. (Local, Amazon S3, Azure Blob Storage, Google Cloud Storage, Dropbox etc...)
+ *
+ * @see https://flysystem.thephpleague.com
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Chris Simpson
+ */
+class Flysystem implements Persister, Stringable
+{
+    /**
+     * The extension to give files created as part of a persistable's save history.
+     *
+     * @var string
+     */
+    public const HISTORY_EXT = 'old';
+
+    /**
+     * The path to the model file on the filesystem.
+     *
+     * @var string
+     */
+    protected $path;
+
+    /**
+     * The filesystem implementation providing access to your backend storage.
+     *
+     * @var \League\Flysystem\FilesystemInterface
+     */
+    protected $filesystem;
+
+    /**
+     * Should we keep a history of past saves?
+     *
+     * @var bool
+     */
+    protected $history;
+
+    /**
+     * The serializer used to convert to and from serial format.
+     *
+     * @var \Rubix\ML\Persisters\Serializers\Serializer
+     */
+    protected $serializer;
+
+    /**
+     * Local
+     *
+     * Shortcut to return a Flysystem Persister backed by the Local filesystem
+     *
+     * @see https://flysystem.thephpleague.com/v1/docs/adapter/local/
+     *
+     * @param string $path The absolute path to the file on the filesystem.
+     * @param bool $history
+     * @param \Rubix\ML\Persisters\Serializers\Serializer|null $serializer
+     * @return \Rubix\ML\Persisters\Flysystem
+     */
+    public static function local(string $path, bool $history = false, ?Serializer $serializer = null) : Flysystem
+    {
+        $filesystem = new League\Flysystem\Filesystem(new Local(dirname($path)));
+
+        return new Flysystem(basename($path), $filesystem, $history, $serializer);
+    }
+
+    /**
+     * FTP
+     *
+     * Shortcut to return a Flysystem Persister backed by an FTP Server
+     *
+     * @see https://flysystem.thephpleague.com/v1/docs/adapter/ftp/
+     *
+     * @param string $path
+     * @param array<mixed> $config Configuration settings for the FTP adapter.
+     * @param bool $history
+     * @param \Rubix\ML\Persisters\Serializers\Serializer|null $serializer
+     * @return \Rubix\ML\Persisters\Flysystem
+     */
+    public static function ftp(string $path, array $config, bool $history = false, ?Serializer $serializer = null) : Flysystem
+    {
+        $filesystem = new League\Flysystem\Filesystem(new Ftp($config));
+
+        return new Flysystem($path, $filesystem, $history, $serializer);
+    }
+
+    /**
+     * @param string $path
+     * @param FilesystemInterface $filesystem
+     * @param bool $history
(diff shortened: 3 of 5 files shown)


src/Persisters/Flysystem.php
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Save the persistable model.
+     *
+     * @param \Rubix\ML\Persistable $persistable
+     * @throws \RuntimeException
+     */
+    public function save(Persistable $persistable) : void
+    {
+        $content = $this->serializer->serialize($persistable);
+        $success = $this->filesystem->put($this->path, $content->data());
```
andrewdalpino: Since Encoding objects are 'Stringable' you might not need to called `data()` here ... not sure because I don't know how Flysystem works under the hood but I think you can just pass the Encoding directly as a string type-hinted argument and it will be converted to string under the hood.

src/Persisters/Flysystem.php
```diff
@@ -0,0 +1,124 @@
...
+
+        if ($data->bytes() === 0) {
+            throw new RuntimeException('File does not contain any data'
+                . " (Path: '{$this->path}', Filesystem: '{$this->getFlysystemAdapterName()}')");
+        }
+
+        return $this->serializer->unserialize($data);
+    }
+
+    protected function getFlysystemAdapterName() : string
```
andrewdalpino: Will this method only ever be used to get a partial string representation of the object? If so maybe consider inlining this code in the __toString() method?

tests/Persisters/FlysystemTest.php
```diff
@@ -0,0 +1,126 @@
+<?php
```
andrewdalpino, line 1: Nice thorough test!

composer.json
```diff
@@ -35,13 +35,15 @@
...
         "amphp/parallel": "^1.3",
+        "league/flysystem": "^1.1",
         "psr/log": "^1.1",
         "rubix/tensor": "^2.0",
         "symfony/polyfill-mbstring": "^1.0",
         "symfony/polyfill-php80": "^1.17"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "2.16.*",
+        "league/flysystem-memory": "^1.0",
```
andrewdalpino, line 46: Nice catch (importing as a dev dependency rather than a project dependency)!

tests/Persisters/FlysystemTest.php
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * @var \Rubix\ML\Persisters\Flysystem
+     */
+    protected $persister;
+
+    /**
+     * @var string
+     */
+    protected $path = 'test.model';
```
andrewdalpino: Will this value ever change or can this be a class constant?

andrewdalpino (commented):
Looks good @simplechris ... I left some general comments and suggestions otherwise this looks good to go. Can you switch targets to the `0.2.0` branch? Since this is a new feature, I'd like to include it in a new feature release (we plan to release 0.2.0 soon).

If you decide to make changes let me know and I'll do a quick review

src/Persisters/Flysystem.php
```diff
@@ -0,0 +1,124 @@
...
+ * user-provided flysystem instance.
+ *
+ * Flysystem is a filesystem abstraction providing a unified interface for many different
+ * filesystems. (Local, Amazon S3, Azure Blob Storage, Google Cloud Storage, Dropbox etc...)
+ *
+ * @see https://flysystem.thephpleague.com
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Andrew DalPino
```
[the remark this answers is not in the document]
simplechris: Yeah, wasn't sure what the precedent was here. I can credit myself if you're cool with that. 👍

docs/persisters/flysystem.md
```diff
@@ -0,0 +1,75 @@
...
+## Parameters
+| # | Param | Default | Type | Description |
+|---|---|---|---|---|
+| 1 | path | | `string` | The path to the model file on the filesystem. |
+| 2 | filesystem |  | `FilesystemInterface` | The flysystem object providing access to your storage backend |
+| 3 | serializer | `Native` | `Serializer` | The serializer used to convert to and from storage format. |
+
+## Examples
+
+#### Local:
```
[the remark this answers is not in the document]
simplechris: Good spot!

src/Persisters/Flysystem.php
```diff
@@ -0,0 +1,124 @@
...
+    protected $filesystem;
+
+    /**
+     * The serializer used to convert to and from serial format.
+     *
+     * @var \Rubix\ML\Persisters\Serializers\Serializer
+     */
+    protected $serializer;
+
+    /**
```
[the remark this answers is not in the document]
simplechris: Yeah, I liked this idea - I'll take another pass at this in the morning.

The reason I didn't include this in this PR was actually a naming/namespace conundrum: Let me brain-dump my thinking here:

-  I had initially planned to make the `FilesystemInterface` an optional constructor parameter (where we would supply a default instance if the user didn't explicitly pass their own) I assumed the default would resolve to the local filesystem - created via a static convenience method like the one you suggest above.
- However `new Filesystem(...)` would clash with the existing `Filesystem` class in the `Rubix\ML\Persistor` namespace. Logically, aliasing this as `Flysystem` would make sense but then we would clash with _this_ class.
- I then wanted to differentiate  why you would use the `Flysystem` Persister vs the  existing `Filesystem` Persister: Remote filesystems. but aliasing  `League\Flysystem\Filesystem` as `RemoteFilesystem` would be valid however a call to `new RemoteFilesystem(new Local(...));` felt 'contradictory'.
- The obvious solution is to have the convenience method use a fully-qualified namespace, but that looks kinda messy and there is no precedent in this library for doing that.
- Finally: as `league/flysystem-memory` is now a `dev` dependency, the 'core' implementations provided by `league/flysysteem` are local and FTP. I wasn't sure if that's a wide enough range to warrant providing convenience/sugar methods.

Any input to help me break my stalemate on this!?  🤣

src/Persisters/Flysystem.php
```diff
@@ -0,0 +1,124 @@
...
+     * Save the persistable model.
+     *
+     * @param \Rubix\ML\Persistable $persistable
+     * @throws \RuntimeException
+     */
+    public function save(Persistable $persistable) : void
+    {
+        $content = $this->serializer->serialize($persistable);
```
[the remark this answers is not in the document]
simplechris: Good idea. I avoided `$data` to avoid the 'stutter' of `$data->data()` below. but I guess this could be avoided by relying on the explicit string cast (if flysystem doesn't already enforce a type cast to string internally - I will double check this)

src/Persisters/Flysystem.php
simplechris: Yeah, i'll check out what flysystem is doing behind-the-scenes 👍 

**EDIT:** Internally the docblocks hint that $contents should be a string, but no explicit casting happens at a flysystem level. Whilst certain implementations may string cast, it's safest to perform that type conversion _here_ so we can ensure the data arriving at any user provided adapter is of the correct type.

src/Persisters/Flysystem.php
simplechris: Yeah. I like this idea.

src/Persisters/Flysystem.php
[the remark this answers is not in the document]
andrewdalpino: Yes go for it! I'd like to keep track of the authors of the file this way

src/Persisters/Flysystem.php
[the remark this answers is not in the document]
andrewdalpino: LOL naming is always the hardest part

Aliasing sounds the easiest ... what about FlyFilesystem as an alias ... it's like a regular FS but has charisma and good taste?

> Finally: as league/flysystem-memory is now a dev dependency, the 'core' implementations provided by league/flysysteem are local and FTP. I wasn't sure if that's a wide enough range to warrant providing convenience/sugar methods.

Yeah I saw that, ftp and local are fine. It's a nice bit of sugar to show our users we care about their volume of keystrokes

src/Persisters/Flysystem.php
andrewdalpino: Agreed

andrewdalpino:
Oh I just thought of 1 more thing! Do you think a 'history' mode would be cool for this Persister? Kinda like the one [here](https://github.com/RubixML/RubixML/blob/master/src/Persisters/Filesystem.php#L84).

simplechris:
@andrewdalpino I've made some changes: [(diff)](https://github.com/RubixML/RubixML/compare/f285437977b7f887c542d5fece79032de438adb0..06115e4). 

###### SUMMARY:
- This PR now targets the `0.2.0` branch.
- `history` functionality added.
- `Flysystem::local('/path/to/model.example')` shortcut method added.
- `Flysystem::ftp('/path/to/model.example', $config)` shortcut method added.
-  Misc changes as suggested in the PR review above.

tests/Persisters/FlysystemTest.php
simplechris: Have now also added tests for the newly introduced `history` mode 👍

simplechris:
Final observations:

#### WINDOWS:
I think my assumption that the root of the filesystem is `/` may not true for every environment? 

I've only ever developed on unix-like systems so admit I know very little about windows (but assume we're supporting it in this lib?) . This might not even be a problem?
I _suppose_ we can do our best to set a logical default, but if a specific user is running an environment where the `PATH_TO_ROOT` is incorrectly set they could just provide the correct value by bypassing the shortcut methods and setting their filesystem route directly:

```php
$filesystem = new \League\Flysystem\Filesystem(
    new Local(/** whatever **/)
);

$persister = new Flysystem(__DIR__ . '/test.model', $filesystem);
```

Or would we need to conditionally in a constant based on the environment:

```php
namespace
{
    if (stripos(PHP_OS, 'WIN') === 0) {
        define('PATH_TO_ROOT', 'c:\\');
    } else {
        define('PATH_TO_ROOT', '/');
    }
}
```

or we could expect the user to define this constant in their own code somewhere, to hint where their paths are relative to

```php
define('PATH_TO_ROOT', '/Users/chris/Datasets/'); 
```

Please let me know if you have any thoughts! I may be overthinking this.

---

#### `FLYSYSTEM` PERSISTER AND `FILESYSTEM` PERSISTER FEATURE PARITY:

Assuming we get a satisfactory solution to the above `PATH_TO_ROOT` question and now that `history` is supported by the `Flysystem` Persister, I _think_ `Filesystem` Persister could be technically refactored to:

```php
<?php

namespace Rubix\ML\Persisters;

use League;
use League\Flysystem\Adapter\Local;
use Stringable;


/**
 * Filesystem
 *
 * Filesystems are local or remote storage drives that are organized by files
 * and folders. The filesystem persister serializes models to a file at a
 * user-specified path.
 *
 * @category    Machine Learning
 * @package     Rubix/ML
 * @author      Andrew DalPino
 */
class Filesystem extends Flysystem implements Persister, Stringable
{
    /**
     * @param string $path
     * @param bool $history
     * @param \Rubix\ML\Persisters\Serializers\Serializer|null $serializer
     */
    public function __construct(string $path, bool $history = false, ?Serializer $serializer = null)
    {
        $filesystem = new League\Flysystem\Filesystem(new Local(PATH_TO_ROOT));

        parent::__construct($path, $filesystem, $history, $serializer);
    }
}
```

Do you think that's possible/worth it?

andrewdalpino:
I'm with you on refactoring the Filesystem persister ... maybe even adding a deprecation annotation and potentially even throwing a deprecation warning in the future. Maybe we'll drop it for Flysystem in 1.0.0. Did you want to do the refactoring as part of this PR or another one?

I don't really understand the first issue. I use Windows so I can test if need be. The PR looks good as long as we don't have an issue with Windows.

andrewdalpino (approved):

simplechris:
> I'm with you on refactoring the Filesystem persister ... maybe even adding a deprecation annotation and potentially even throwing a deprecation warning in the future. Maybe we'll drop it for Flysystem in 1.0.0. Did you want to do the refactoring as part of this PR or another one?

This refactor can totally wait for a future PR, but like you say - should probably come before `v1.0.0`, sooner if you want to follow a deprecation schedule 👍 

---

> I don't really understand the first issue. I use Windows so I can test if need be. The PR looks good as long as we don't have an issue with Windows.

Gotcha! I did a poor job at describing the problem 🤣 . The issue is the `Flysystem::local()` shortcut. It currently doesn't allow of a way to configure the 'root' of the filesystem (the only required argument to the `League\Flysystem\Adapter\Local()` class. This is the location that `$path` is defined relative to) 

Assuming that users are providing the full (absolute) path to their file on the filesystem get then this can just default to the root of the filesystem (which on *nix operating systems is `/`, and upon further research i've discovered that on windows is `\` (or could be something like `D:\` for a system wide reference to a different disk)

So i think we should accept an additional parameter for this helper function: `$root`.  If the user is defining their `$path` variable relative to a specific directory they can provide that reference to the `$root` variable. If a specific value is not provided it should default to the filesystem root (which is system specific, but can be inferred by the `PHP_OS` constant)

How about something like this: 

```php

public static function local(string $path, bool $history = false, ?Serializer $serializer = null, $root = null) : Flysystem
{

    if (!$root) {
        // No $root - $path is an _absolute_ path.
        // An absolute path is provided relative to the filesystem root:
        //    - on a Unix file system it's `/`
        //    - on Windows it's either `\` (for the current disk) or `D:\` (system-wide).
        $root = (stripos(PHP_OS, 'WIN') === 0) ? '\\' : '/';  
    }

    $filesystem = new League\Flysystem\Filesystem(new Local($root));

    return new Flysystem($path, $filesystem, $history, $serializer);
}

```

I'll go ahead and push this change now, but feel free to suggest edits if you think that this isn't the optimal solution.

simplechris:
Hey @andrewdalpino, 

the build was initially failing on windows (https://github.com/RubixML/RubixML/runs/1081862469) due to `ext-fileinfo` been missing in the environment of the windows box (despite been enabled by default since php5.3? 🤷‍♂️ ) I've added it explicitly here: https://github.com/RubixML/RubixML/pull/107/commits/a251f8accbe4cee8c3faff5ef313daa4c557c787 but now i'm getting static analysis failures relating to the svm extension. I may need you to help me debug this further (It's unrelated to the latest code changes on this branch) 👍

andrewdalpino:
Gotcha @simplechris

I have a Windows box so I can test. I thought the windows shells worked fine with Unix style paths though.

Annotation 2020-09-07 150142

Powershell seems to recognize the Unix notation and work with it ok. Are we certain it (a constant like you had before) doesn't work on Windows?

> the build was initially failing on windows (https://github.com/RubixML/RubixML/runs/1081862469) due to ext-fileinfo been missing in the environment of the windows box (despite been enabled by default since php5.3? 🤷‍♂️ ) I've added it explicitly here: a251f8a but now i'm getting static analysis failures relating to the svm extension. I may need you to help me debug this further (It's unrelated to the latest code changes on this branch) 👍

I added Windows to the CI pipeline yesterday just FYI. The instances they seem to spin up aren't the same as Ubuntu or MacOS, for example svm extension is not working. Not sure if you know of a way to fix this but I can always remove Windows from the pipeline as a temporary fix. Let me know!

P.S. Removed windows for the time being

simplechris:
I've found a solution to this.

This implementation works on all platforms:

```php
/**
 * Local
 *
 * Shortcut to return a Flysystem Persister backed by the Local filesystem
 *
 * @see https://flysystem.thephpleague.com/v1/docs/adapter/local/
 *
 * @param string $path The absolute path to the file on the filesystem.
 * @param bool $history
 * @param \Rubix\ML\Persisters\Serializers\Serializer|null $serializer
 * @return \Rubix\ML\Persisters\Flysystem
 */
public static function local(string $path, bool $history = false, ?Serializer $serializer = null) : Flysystem
{
    $filesystem = new League\Flysystem\Filesystem(new Local(dirname($path)));

    return new Flysystem(basename($path), $filesystem, $history, $serializer);
}
```
Heres a link to the builds passing (on windows too) https://github.com/simplechris/RubixML/runs/1087516451?check_suite_focus=true

(I explicitly ignored svm dependent classes from static analysis step to bypass the windows boxes in CI not installing svm extension correctly) 

I'll cherry pick this over into this branch shortly.

andrewdalpino:
Nice work @simplechris!

One last question before we merge this ... would there be any reason that some obscure Flysystem adapter would not work with this setup? For example, I'm thinking if it didn't support renaming or something like that. I still say we go with it even if there are potential issues with certain adapters but I'd just like us to have an idea beforehand.

simplechris:
> would there be any reason that some obscure Flysystem adapter would not work with this setup? For example, I'm thinking if it didn't support renaming or something like that. I still say we go with it even if there are potential issues with certain adapters but I'd just like us to have an idea beforehand.

I don't think so, any adapter has to implement the `AdapterInterface` so we can rely on that contract. Even if the adapter internally implements `rename` in two individual operations (as a write followed by a delete) then we'd still have the desired behaviour. The interface defines which exceptions can theoretically be thrown in the adapters implementation, so I could add even more explicit exception handling around that call if wanted.

andrewdalpino:
Nah, that's not necessary - like I said just something to have thought about beforehand

Thanks again for the PR @simplechris!

PR #109, opened 2020-09-10
Upgrade to Flysystem 2.0

andrewdalpino:
This upgrades the Flysystem persister to use Flysystem 2.0 (which is currently in beta) under the hood. They do not have nearly half as many adapters available yet but I think this will be a good move in the long run. The "plugin" pattern seems to cause alot of problems and I like the new design of 2.0 so far.

### DIFF
--- a/src/Persisters/Flysystem.php
+++ b/src/Persisters/Flysystem.php
@@ -9,9 +9,9 @@ use Rubix\ML\Persisters\Serializers\Native;
 use Rubix\ML\Persisters\Serializers\Serializer;
 use League\Flysystem\Filesystem;
 use League\Flysystem\FilesystemOperator;
+use League\Flysystem\FilesystemException;
 use RuntimeException;
 use Stringable;
-use Exception;
 
 /**
  * Flysystem
@@ -104,7 +104,7 @@ class Flysystem implements Persister, Stringable
 
             try {
                 $this->filesystem->move($this->path, $filename);
-            } catch (Exception $e) {
+            } catch (FilesystemException $e) {
                 throw new RuntimeException("Failed to create history file '$filename'.");
             }
         }
@@ -113,7 +113,7 @@ class Flysystem implements Persister, Stringable
 
         try {
             $this->filesystem->write($this->path, $encoding);
-        } catch (Exception $e) {
+        } catch (FilesystemException $e) {
             throw new RuntimeException('Could not write to filesystem.');
         }
     }
@@ -130,7 +130,13 @@ class Flysystem implements Persister, Stringable
             throw new RuntimeException("File does not exist at {$this->path}.");
         }
 
-        $encoding = new Encoding($this->filesystem->read($this->path) ?: '');
+        try {
+            $data = $this->filesystem->read($this->path);
+        } catch (FilesystemException $e) {
+            throw new RuntimeException("Error reading data from {$this->path}.");
+        }
+
+        $encoding = new Encoding($data);
 
         if ($encoding->bytes() === 0) {
             throw new RuntimeException("File at {$this->path} does not contain any data.");
--- a/src/Persisters/Flysystem.php
+++ b/src/Persisters/Flysystem.php
@@ -2,32 +2,31 @@
 
 namespace Rubix\ML\Persisters;
 
-use League;
-use League\Flysystem\Adapter\Local;
-use League\Flysystem\Adapter\Ftp;
-use League\Flysystem\FilesystemInterface;
 use Rubix\ML\Encoding;
 use Rubix\ML\Persistable;
 use Rubix\ML\Other\Helpers\Params;
 use Rubix\ML\Persisters\Serializers\Native;
 use Rubix\ML\Persisters\Serializers\Serializer;
+use League\Flysystem\Filesystem;
+use League\Flysystem\FilesystemOperator;
 use RuntimeException;
 use Stringable;
+use Exception;
 
 /**
  * Flysystem
  *
- * The flysystem persister serializes models to a file at a user-specified path, using a
- * user-provided flysystem instance.
+ * Flysystem is a filesystem library providing a unified storage interface and abstraction layer.
+ * It enables access to many different storage backends such as Local, Amazon S3, FTP, and more.
  *
- * Flysystem is a filesystem abstraction providing a unified interface for many different
- * filesystems. (Local, Amazon S3, Azure Blob Storage, Google Cloud Storage, Dropbox etc...)
+ * > **Note:** The Flysystem persister is designed to work with Flysystem version 2.0.
  *
  * @see https://flysystem.thephpleague.com
  *
  * @category    Machine Learning
  * @package     Rubix/ML
  * @author      Chris Simpson
+ * @author      Andrew DalPino
  */
 class Flysystem implements Persister, Stringable
 {
@@ -48,7 +47,7 @@ class Flysystem implements Persister, Stringable
     /**
      * The filesystem implementation providing access to your backend storage.
      *
-     * @var \League\Flysystem\FilesystemInterface
+     * @var \League\Flysystem\FilesystemOperator
      */
     protected $filesystem;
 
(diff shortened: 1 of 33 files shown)


src/Persisters/Flysystem.php
```diff
@@ -2,32 +2,31 @@
...
-use League;
-use League\Flysystem\Adapter\Local;
-use League\Flysystem\Adapter\Ftp;
-use League\Flysystem\FilesystemInterface;
 use Rubix\ML\Encoding;
 use Rubix\ML\Persistable;
 use Rubix\ML\Other\Helpers\Params;
 use Rubix\ML\Persisters\Serializers\Native;
 use Rubix\ML\Persisters\Serializers\Serializer;
+use League\Flysystem\Filesystem;
```
simplechris, line 10: ℹ️   unused

src/Persisters/Flysystem.php
```diff
@@ -182,12 +152,9 @@ public function load() : Persistable
...
-
-        if (method_exists($this->filesystem, 'getAdapter')) {
-            $filesystem .= ' <' . Params::toString($this->filesystem->getAdapter()) . '>';
-        }
-
-        return "{$filesystem} (path: '{$this->path}', serializer: {$this->serializer})";
+        return "Flysystem (path: {$this->path}, filesystem: "
+            . Params::toString($this->filesystem) . ', history: '
+            . Params::toString($this->history) . ', serializer: '
+            . "{$this->serializer})";
```
simplechris, line 158: I think it's more informative to have the _adapter_ (`League\Flysystem\AdapterInterface`) used not the `FilesystemOperator` wrapper class. As `FilesystemOperator` (and its predecessor `FilesystemInterface`) doesn't enforce access to it, I had the runtime reflection to check.

My aim was to have the string output be something along the lines of:

```
Flysystem (path: path/to/test.persistable, filesystem: InMemory, history: false, serializer: Native)
```

rather than: 
```
Flysystem (path: path/to/test.persistable, filesystem: Filesystem, history: false, serializer: Native)
```

tests/Persisters/FilesystemTest.php
```diff
@@ -15,6 +15,13 @@
  */
 class FilesystemTest extends TestCase
 {
+    /**
+     * The path to the test file.
+     *
+     * @var string
+     */
+    const PATH = __DIR__ . '/test.persistable';
```
simplechris, line 23: will `__DIR__` always have write permissions on all systems? 

EDIT: It probably will thinking about it, but you can always safely write to the temp dir returned by `sys_get_temp_dir()`, however that would mean using a class property and not a constant, I think. meh.

simplechris (commented):
This is great! left a few comments inline.

I much prefer the new interfaces that flysystem v2 provides. I would have implemented #107 using this version had it been out of beta and in release-candidate phase (but i'm fairly confident that the core interfaces will not change between now and general release)

I also agree with your decision to remove the static convenience/factory methods from the `Rubix\ML\Persisters\Flysystem` class.

👍

src/Persisters/Flysystem.php
```diff
@@ -162,17 +126,23 @@ public function save(Persistable $persistable) : void
...
+            throw new RuntimeException("Error reading data from {$this->path}.");
         }
 
-        $data = new Encoding($this->filesystem->read($this->path) ?: '');
+        $encoding = new Encoding($data);
 
-        if ($data->bytes() === 0) {
-            throw new RuntimeException("File does not contain any data. {$this}");
+        if ($encoding->bytes() === 0) {
+            throw new RuntimeException("File at {$this->path} does not contain any data.");
```
simplechris, line 142: ❓  **Question:** As we are throwing a `RuntimeException` when trying to _read_ an empty encodings, should we also raise a similar exception in the `save` method if trying to _write_ an empty encoding? 🤔

src/Persisters/Flysystem.php
```diff
@@ -66,91 +65,56 @@ class Flysystem implements Persister, Stringable
...
             try {
-                if (!$this->filesystem->rename($this->path, $filename)) {
-                    throw new RuntimeException("Failed to create history file: '{$filename}' {$this}");
-                }
-            } catch (League\Flysystem\Exception $e) {
-                throw new RuntimeException("Failed to create history file: '{$filename}' {$this}");
+                $this->filesystem->move($this->path, $filename);
+            } catch (FilesystemException $e) {
+                throw new RuntimeException("Failed to create history file '$filename'.");
             }
```
simplechris, line 109: I've just noticed that `fileExists` implementations can throw exceptions too. We should move the `while` statement into the `try` block beneath.

src/Persisters/Flysystem.php
```diff
@@ -162,17 +126,23 @@ public function save(Persistable $persistable) : void
...
-            throw new RuntimeException("File does not exist in filesystem. {$this}");
+        if (!$this->filesystem->fileExists($this->path)) {
+            throw new RuntimeException("File does not exist at {$this->path}.");
+        }
+
+        try {
+            $data = $this->filesystem->read($this->path);
+        } catch (FilesystemException $e) {
+            throw new RuntimeException("Error reading data from {$this->path}.");
         }
```
simplechris, line 137: As above, let's start the `try` block at start of this method so that anything raised from `fileExists` is also handled.

simplechris (commented):
My favourite improvement to the design of Flysystem v2 is that error states are always indicated by the throwing of an exception (not the mix of success booleans and exceptions that v1.x used). This means any invocation of methods on the `FilesystemOperator` should probably be wrapped in a try/catch.

My previous implementation also would have suffered from this issue - that `fileExists` (previously `has`) could throw unhandled errors.

src/Persisters/Flysystem.php
andrewdalpino: Nice catch!

src/Persisters/Flysystem.php
andrewdalpino: Yeah I liked the way you had it before, will probably put back

tests/Persisters/FilesystemTest.php
andrewdalpino: Interesting, I think I'll try that

Thanks!

src/Persisters/Flysystem.php
andrewdalpino: Yes, we should probably be consistent, nice catch

src/Persisters/Flysystem.php
andrewdalpino: Hmmmmm thanks for pointing that

tests/Persisters/FilesystemTest.php
simplechris: The cleanest solution I found for this last week was defining a constant in the global scope in constants.php so you could use `TMP_DIR` in place of `__DIR__`


See: https://github.com/RubixML/RubixML/pull/106#discussion_r481751711

andrewdalpino:
Thanks for the review @simplechris! You got to it before I could even give you a heads up (I did this late last night - ignore the mess)

What do you think about using 2.0 vs 1.0 to start out with? I wonder if we should hit them up and see if they can give us an estimate as to when they'll have a stable API.

> I also agree with your decision to remove the static convenience/factory methods from the Rubix\ML\Persisters\Flysystem class.

I realized that some of the adapters required alot of arguments (like FTP, and S3) and thought we'd think about either implementing factory methods or creating separate objects (persisters) that extend the Flysystem base class (like the current Filesystem persister). I figured we didn't need to decide right away to get 0.2.0 released. Thoughts?

src/Persisters/Flysystem.php
simplechris: I admit it looked a little messy, I did abstract this logic into a helper method in a branch i have locally: 

```php
Params::property(object $subject, string $property, string $fallback = ''): mixed;
```
so could be used as:

```php
Params::property($this->filesystem, 'adapter')
```

which internally uses reflection if the property is `protected` or `private`, the fallback value is used if the property doesn't exist on the subject.

Can push it up to a branch later and you can decide if you like it or not 👍

Issue #108, opened 2020-09-04
Integrate persistence subsystem with all the things!!!


andrewdalpino:
Right now, the persistence abstraction is limited to Estimators, however, think of how cool it would be to have them work with *everything*. This could be one of the major enhancements we deliver in the next major release (1.0.0). This ticket is to discuss the design of the new persistence subsystem. The idea is to be able to use Persisters as a one-stop-shop.

The relevant abstractions we current have to work with are:

- [Persisters](https://docs.rubixml.com/en/latest/persisters/api.html) - save and load chunks of encoded data
- [Serailizers](https://github.com/RubixML/RubixML/blob/master/src/Persisters/Serializers/Serializer.php) - turn objects into chunks of encoded data
- [Persistable objects](https://docs.rubixml.com/en/latest/persistable.html) - objects whitelisted for persistence
- [Encoding](https://github.com/RubixML/RubixML/blob/master/src/Encoding.php) - a low-level value object used throughout Rubix to wrap encoded data
- [CSV](https://docs.rubixml.com/en/latest/extractors/csv.html), [JSON](https://docs.rubixml.com/en/latest/extractors/json.html), [NDJSON](https://docs.rubixml.com/en/latest/extractors/ndjson.html) extractors - since they deal with loading encoded data from a file
- [Transformers](https://docs.rubixml.com/en/latest/transformers/api.html) (Stateful and Elastic) - objects we'd like to persist using the persistence subsystem but have not implemented yet

Backwards compatibility breaks are fine since this targets the next major release.

Some place to start ...

We can probably replace any string `$path` arguments on Encoding, CSV, NDJSON, and JSON with a Persister interface instead. It's a little more to type out in the constructor but it's a powerful abstraction. Imagine loading CSV datasets from a Redis database or JSON dataset from a MongoDB database. Or saving a cross-validation report directly to a file on an S3 server in  JSON format.

The Persistable interface is actually just a whitelist marker for "any object that can safely be persisted". It shouldn't be too hard to just have Transformers 'implement' the interface and hop on the persistence train.

andrewdalpino:
I think something else that we should incorporate into this one-persistence-subsystem-to-rule-them-all is a way to store meta data on the persisted object. For example, the method used for serialization, the version number of the class (See https://github.com/RubixML/RubixML/issues/115), and maybe even a checksum.

CC @simplechris @ChristophWurst

andrewdalpino:
[...]

PR #110, opened 2020-09-10
Revert "Flysystem Persister"

andrewdalpino:
We agreed (although hesitant) to wait until the next release (hopefully no more) to introduce Flysystem as A. We like the idea of using Flysystem 2.0 but B. They are still in beta currently. We hope that by waiting we can mitigate having to make breaking changes to the Flysystem persister if Flysystem decides to change their API between now and their first stable 2.0 release.

CC: @simplechris


simplechris:
Yeah, although I'm keen to see this enhancement get into `master`, delaying this to the `0.3.0` release this is the right call, imo.

- Flysystem 2.x will have a far longer lifespan, and the design is objectively better/cleaner.
- Considering a `2.0-beta3` release already out, it's just a matter of time before they optionally move to RC phase or directly to a stable release 🤞.
- Although it's unlikely that the core interfaces will change in Flysystem 2.x (at this stage) crazier things have happened in software development. A little patience now could pay dividends in the future for us and avoid introducing any breaking API changes.
- In the meantime I can open something against the upcoming `0.3.0` branch, that also includes the enhancements spotted in #109.
- Also related: I'm making decent progress on #108 and hope to have a design proposal soon. We can use as starting point for designing a more fully-featured persistence subsystem. In turn, the existing interfaces can take advantage of this internally as we progress (in time) to the `1.0` release.

PR #112, opened 2020-09-11
Avoid persisting empty encoding objects

### LINKED ISSUE #111: Avoid persisting empty encoding objects
❓  **Question:** As we are throwing a `RuntimeException` when trying to _read_ an empty encodings, should we also raise a similar exception in the `save` method if trying to _write_ an empty encoding? 🤔

_Originally posted by @simplechris in https://github.com/RubixML/RubixML/pull/109#discussion_r486381555_

simplechris:
###### SUMMARY:

Pulled this out of the review of #109, figured this could still make it in to `0.2.0`?

---

- Closes #111


andrewdalpino (approved):
Nice thank you!

PR #114, opened 2020-09-21
cleanup

simplechris:
#### SUMMARY:

Collated a few minor cleanups that I had stashed locally (docblock fixes, implicit type conversions etc)

### DIFF
--- a/src/Other/Tokenizers/SkipGram.php
+++ b/src/Other/Tokenizers/SkipGram.php
@@ -85,7 +85,7 @@ class SkipGram implements Tokenizer, Stringable
      * Tokenize a blob of text.
      *
      * @param string $string
-     * @return string[]
+     * @return list<string>
      */
     public function tokenize(string $string) : array
     {
--- a/src/Other/Tokenizers/Tokenizer.php
+++ b/src/Other/Tokenizers/Tokenizer.php
@@ -8,7 +8,7 @@ interface Tokenizer
      * Tokenize a blob of text.
      *
      * @param string $string
-     * @return string[]
+     * @return list<string>
      */
     public function tokenize(string $string) : array;
 
--- a/src/Classifiers/MultilayerPerceptron.php
+++ b/src/Classifiers/MultilayerPerceptron.php
@@ -177,7 +177,6 @@ class MultilayerPerceptron implements Estimator, Learner, Online, Probabilistic,
      * @param float $holdOut
      * @param \Rubix\ML\NeuralNet\CostFunctions\ClassificationLoss|null $costFn
      * @param \Rubix\ML\CrossValidation\Metrics\Metric|null $metric
-     * @param float $holdOut
      * @throws \InvalidArgumentException
      */
     public function __construct(
--- a/src/NeuralNet/Optimizers/AdaGrad.php
+++ b/src/NeuralNet/Optimizers/AdaGrad.php
@@ -40,7 +40,7 @@ class AdaGrad implements Optimizer, Adaptive, Stringable
      *
      * @var \Tensor\Tensor[]
      */
-    protected $cache;
+    protected $cache = [];
 
     /**
      * @param float $rate
--- a/src/Other/Tokenizers/NGram.php
+++ b/src/Other/Tokenizers/NGram.php
@@ -77,10 +77,10 @@ class NGram implements Tokenizer, Stringable
     }
 
     /**
-     * Tokenize a block of text.
+     * Tokenize a blob of text.
      *
      * @param string $string
-     * @return string[]
+     * @return list<string>
      */
     public function tokenize(string $string) : array
     {
--- a/src/Other/Tokenizers/Whitespace.php
+++ b/src/Other/Tokenizers/Whitespace.php
@@ -41,7 +41,7 @@ class Whitespace implements Tokenizer, Stringable
      * Tokenize a blob of text.
      *
      * @param string $string
-     * @return string[]
+     * @return list<string>
      */
     public function tokenize(string $string) : array
     {
--- a/src/Other/Tokenizers/Word.php
+++ b/src/Other/Tokenizers/Word.php
@@ -26,7 +26,7 @@ class Word implements Tokenizer, Stringable
      * Tokenize a blob of text.
      *
      * @param string $string
-     * @return string[]
+     * @return list<string>
      */
     public function tokenize(string $string) : array
     {
--- a/src/Persisters/Serializers/Igbinary.php
+++ b/src/Persisters/Serializers/Igbinary.php
@@ -44,12 +44,12 @@ class Igbinary implements Serializer, Stringable
     /**
      * Unserialize a persistable object and return it.
      *
-     * @param \Rubix\ML\Encoding $data
+     * @param \Rubix\ML\Encoding $encoding
      * @return \Rubix\ML\Persistable
      */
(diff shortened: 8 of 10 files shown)


src/Other/Tokenizers/SkipGram.php
```diff
@@ -81,12 +81,6 @@ public function __construct(int $n = 2, int $skip = 2, ?Word $wordTokenizer = nu
         $this->wordTokenizer = $wordTokenizer ?? new Word();
     }
 
-    /**
-     * Tokenize a blob of text.
-     *
-     * @param string $string
-     * @return string[]
```
andrewdalpino, line 88: Each method requires a docblock

src/Other/Tokenizers/Tokenizer.php
```diff
@@ -8,7 +8,7 @@ interface Tokenizer
      * Tokenize a blob of text.
      *
      * @param string $string
-     * @return string[]
+     * @return list<string>
```
andrewdalpino, line 11: What does the list annotation mean?

andrewdalpino (changes requested):
Looks good nice catches but some changes are not consistent with out coding style

src/Other/Tokenizers/SkipGram.php
simplechris: won't this now inherit the docblock defined in the interface `Rubix\ML\Other\Tokenizers::tokenize()`, or do you need an explicit 

```php
/**
 * {@inheritdoc}
 */
```

andrewdalpino: I'm not familiar with {@inheritdoc} but I use these docblocks as a summary of the inputs and outputs of the method. It looks like that information just vanished.

andrewdalpino: Btw we don't use PHPdoc, I use those docblocks for my own reference

https://docs.phpdoc.org/latest/guides/inheritance.html

andrewdalpino: We don't use PHPdoc and I use these docblocks as a reference when I'm programming. If we delete them and put {@inheritdoc} instead then, to me, the information just vanishes. Am I doing it wrong? Is PHPStan compatible with {@inheritdoc} tags?

src/Other/Tokenizers/Tokenizer.php
simplechris: I added this in https://github.com/RubixML/RubixML/pull/114/commits/47510dc78d6e01724adbaa5431edda981869824e but to expand in some more detail:

`list<string>` is a way of enforcing that the return type will not be an associative array, rather a list of strings (a list will have numeric keys) This is important because some implementations like `NGram`, `SkipGram` work on the assumption that the key of the array will be an integer, this annotation enforces this assumption. This is why the removal of the non-essential docblocks is actually meaningful in this scenario. 

**EDIT:** This is why the removal of the non-essential docblocks is actually meaningful in this scenario: so that the annotation defined on the interface is read by static analyzers and not overwritten by the inclusion of a docblock on the implementing class/child (that adds no further detail)

andrewdalpino: Interesting! Thanks for that. We mostly use sequential (non-associative) arrays so this annotation may be very useful across the board.

src/Other/Tokenizers/SkipGram.php
simplechris: I expanded as to why the removal of these non-essential docblocks is actually meaningful for static analysis in this scenario in this comment: https://github.com/RubixML/RubixML/pull/114#discussion_r492344856

I'm not advocating for the use of inheritdoc tags, but if the coding standard enforces a docblock then `{@inheritdoc}` would satisfy that condition whilst also introducing increased type-safety.

src/Other/Tokenizers/Tokenizer.php
simplechris: I'm glad you think so, because this was a PR that was 'testing the water' I've become to rely on `list` annotations in my day-job, and have a larger PR related to this stashed away 🤣

src/Other/Tokenizers/SkipGram.php
andrewdalpino: I still don't get it ... considering I look at these docblocks when I'm programming to understand the inputs and outputs, types, and general purpose of the thing - how do I get that info into my brain/short-term memory if all I see is "{@inheritdoc}" in my IDE? It seems that I would have to go hunting for the information that I would normally have up-front. Am I missing the fact that my IDE will do the work for me?

src/Other/Tokenizers/Tokenizer.php
andrewdalpino: Awesome

src/Other/Tokenizers/SkipGram.php
simplechris: Yep. I agree. I also often naturally use docblocks over IDE magic. 

My only reason to remove this docblock as part of this PR is to allow the `@return list<string>` annotation I added to the `Tokenizer` interface to be used by static analyzers like phpstan/psalm:

docblocks naturally inherit from their parent for machines - (parsers use `ReflectionClass::getDocComment()` that naturally traverses to the parent definition). This is why the additional type-safety of introducing `list<string>` is only meaningful when static analyzers can access those annotations. (whilst most projects don't use phpdoc for documentation the phpdoc standard is used by static analyzers to read the annotations added to code) Removing this docblock means that static analysis will now default to the docblock that is available in the interface.

I wasn't really advocating for including `{@inheritdoc}` (personally, I don't like them) I was more suggesting a solution that allowed for this method to have a docblock, but an also inherit the improvement of annotating the return type as a `list<string>`, In previous PRs you have shown an aversion for non-essential/self-evident comments so was incorrectly pre-empting your preference here 🤣.

We could equally just retain these docblocks but _also_ update their return type annotations to be `list<string>`, (or, investigate a `php-cs-fixer` configuration that will enforce this standard and automatically add these in if missing)

andrewdalpino: I wasn't making any statements in the above comment ... I am genuinely confused as to what the purpose/benefit of docblock inheritance is. Will the IDE automatically "inject" the docblocks for me? If so, that would make more sense to me. If not, then let's leave the docblocks, as I find them rather very useful, and just change the annotations individually.

simplechris: Gotcha - I guess inheritdoc just tries to apply a DRY convention to avoid having to update docblocks all over a codebase instead of just on the parent: but it comes at the significant expense of human readability. I'll add them back in with updated return types. 

Anecdotally, my IDE compiles them and displays them like this:

PHPStorm - Screenshot 09:39am - chriss@local


This little discussion here got me thinking about how docblocks inhabit an awkward space between 'comments' and 'code' (they contain both arbitrary comments aimed at humans AND also meaningful annotations/code aimed at machines) Once annotations [[rfc:accepted](https://wiki.php.net/rfc/shorter_attribute_syntax_change)] land in PHP8 annotations will be supported directly at language level and not have to be parsed out of comment blocks, so maybe clearer conventions will emerge then (Maybe: include specific static analysis metadata on the parent using annotations, liberally include helpful explanation in the doc comments wherever you can) 👍

andrewdalpino (approved):

andrewdalpino:
Thank you!

PR #116, opened 2020-09-22
Exceptions

simplechris:
#### SUMMARY:

This PR introduces the `Rubix\ML\Exceptions` namespace.

Existing exceptions in the global namespace  are now thrown as the equivalent exception in the `Rubix\ML\Exceptions` namespace (e.g. `\InvalidArgumentException` is now `\Rubix\ML\Exceptions\InvalidArgumentException`). These exceptions all implement the `RubixMLException` interface which can now act as a generic flag to catch things thrown from within the library.

This PR introduces no BC breaks: if existing user-land code is catching `\RuntimeException`, `\InvalidArgumentException` etc then this is still valid as the new Exceptions also extend them. This paves the way for more expressive and granular error types in the future, and perhaps a larger overhaul of the exceptions hierarchy in 1.0 or beyond. 

I've purposely kept the scope of this PR targeted to the existing exceptions to avoid introducing too many new changes in a single leap.

### DIFF
--- a/src/AnomalyDetectors/GaussianMLE.php
+++ b/src/AnomalyDetectors/GaussianMLE.php
@@ -19,8 +19,8 @@ use Rubix\ML\Other\Traits\PredictsSingle;
 use Rubix\ML\Specifications\DatasetIsNotEmpty;
 use Rubix\ML\Specifications\DatasetHasDimensionality;
 use Rubix\ML\Specifications\SamplesAreCompatibleWithEstimator;
-use InvalidArgumentException;
-use RuntimeException;
+use Rubix\ML\Exceptions\InvalidArgumentException;
+use Rubix\ML\Exceptions\RuntimeException;
 use Stringable;
 
 use function Rubix\ML\warn_deprecated;
@@ -90,7 +90,7 @@ class GaussianMLE implements Estimator, Learner, Online, Scoring, Ranking, Persi
 
     /**
      * @param float $contamination
-     * @throws \InvalidArgumentException
+     * @throws \Rubix\ML\Exceptions\InvalidArgumentException
      */
     public function __construct(float $contamination = 0.1)
     {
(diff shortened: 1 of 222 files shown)


andrewdalpino:
Hey @simplechris I'm wondering what your vision was for the next couple steps. How does this serve as the basis for our exception system in  the futrue?

andrewdalpino:
Ok took some time to let this marinate (visualize the path forward) but I think this is it ... the only thing that is still a toss up for me is the name of our exception interface. Should it be \Rubix\ML\Exceptions\RubixMLException or just \Rubix\ML\Exceptions\Exception?

Also can you pull the new 0.3.0 changes so we can merge via Github?

simplechris (commented):
@andrewdalpino I've re-applied these changes on top of the existing changes in the `0.3.0` branch 👍 

As discussed in chat earlier, using `RubixMLException` as the main interface marker may be most suitable. This allows for a generic exception type of `Rubix\ML\Exceptions\Exception`. We can seek more opinions before making the final decision though 👍 

It may be good to define some convention around the naming of exceptions too. The included base exceptions that extend their counterparts in the standard library follow the convention of including the `Exception` suffix. (`RuntimeException`, `InvalidException` etc ...)

If we add any specialist exception types within the library then I don't think that these should *have* to include the suffix, however by convention they should extend one of the base types in the `Rubix\ML\Exceptions` namespace (enforcing the requirement that they implement `RubixMLException` interface). Anything else you can think of? Also are there any documentation updates that need to be made as part of this PR?

andrewdalpino:
Sounds good @simplechris nice work! We should probably provide exception documentation at some point. Do you have something in mind for the docs already?

andrewdalpino:
@simplechris We need to remember to update the Extras repo with these new exceptions as well before we release Extras 0.3.0-beta

PR #118, opened 2020-09-25
fix BallTree edge case

kroky:
We had another infinite loop problem in BallTree. Whenever a dataset contained the same samples repeating more than the max leaf size BallTree::grow was trying to split the same tree over and over again - since all nodes were having the same left and right centeroid values, it kept all the samples in the right subtree and continued trying to split it. I added an exception handling code that terminated the process in that case. It can result with a leaf node containing more than max leaf size but I am not sure what else to do here - all the samples in that leaf are the same, so we don't have a criteria to further split them.

Added a test to cover this use-case. If you run the test without the fix in BallTree class, you will see an infinite loop occurs.

Let me know if you see other solutions here or any problems with this one. Thanks!

### DIFF
--- a/src/Graph/Trees/BallTree.php
+++ b/src/Graph/Trees/BallTree.php
@@ -150,9 +150,14 @@ class BallTree implements BinaryTree, Spatial, Stringable
             if ($right->numRows() > $this->maxLeafSize) {
                 $node = Ball::split($right, $this->kernel);
 
-                $current->attachRight($node);
+                [$nleft, $nright] = $node->groups();
 
-                $stack[] = $node;
+                if ($left->empty() && $nleft->empty() && $right == $nright) {
+                    $current->attachRight(Clique::terminate($right, $this->kernel));
+                } else {
+                    $current->attachRight($node);
+                    $stack[] = $node;
+                }
             } elseif (!$right->empty()) {
                 $current->attachRight(Clique::terminate($right, $this->kernel));
             }
--- a/tests/Graph/Trees/BallTreeTest.php
+++ b/tests/Graph/Trees/BallTreeTest.php
@@ -2,6 +2,7 @@
 
 namespace Rubix\ML\Tests\Graph\Trees;
 
+use Tensor\Matrix;
 use Rubix\ML\Graph\Trees\Tree;
 use Rubix\ML\Graph\Trees\Spatial;
 use Rubix\ML\Graph\Trees\BallTree;
@@ -9,6 +10,7 @@ use Rubix\ML\Graph\Trees\BinaryTree;
 use Rubix\ML\Datasets\Generators\Blob;
 use Rubix\ML\Kernels\Distance\Euclidean;
 use Rubix\ML\Datasets\Generators\Agglomerate;
+use Rubix\ML\Datasets\Labeled;
 use PHPUnit\Framework\TestCase;
 
 /**
@@ -80,6 +82,32 @@ class BallTreeTest extends TestCase
         $this->assertCount(25, $labels);
         $this->assertCount(25, $distances);
 
+        $this->assertCount(1, array_unique($labels));
+    }
+
+    /**
+     * @test
+     */
+    public function growWithRepetitions() : void
+    {
+        $size = 25;
+
+        $samples = Matrix::fill(0.5, $size, 3)->asArray();
+        $dataset = Labeled::quick($samples, array_fill(0, $size, 'test'));
+
+        $this->tree->grow($dataset);
+
+        $this->assertEquals(2, $this->tree->height());
+
+        $sample = $dataset->sample(0);
+
+        [$samples, $labels, $distances] = $this->tree->nearest($sample, 5);
+
+        $this->assertCount(5, $samples);
+        $this->assertCount(5, $labels);
+        $this->assertCount(5, $distances);
+
+        $this->assertCount(1, array_unique($samples, SORT_REGULAR));
         $this->assertCount(1, array_unique($labels));
     }
 


andrewdalpino:
Damn @kroky you are quite the code detective! I will check this out tomorrow I just wanted to say nice work

Does increasing the `maxLeafSize` to a value greater than the number of same samples (assuming these were obscure texts that all had zero-norm vector representations) work fine?

I wonder if this was the root cause of our issue with Cosine kernel - Would that bug have contributed to this bug or was that a separate bug? Either way I still believe the distances returned with zero vector(s) was still "semantically" faulty.

kroky:
Yes, if `maxLeafSize` is greater than the number of same samples then it worked fine as the node is terminated and no more split attempts occur. My example is very dumb currently with about 69 samples containing the word 'test', so they have non-zero vector representations.

I agree about Cosine kernel - the fixes were needed there as we have correct results now for empty vectors. Otherwise, I don't think the fix here would have fixed that as we have a different situation - currently we have empty left subtree with all elements falling into the right subtree (because Ball node puts samples with equal distance in the right subtree). I believe the previous bug had just one element in the left subtree but did not correctly count the distance, so it kept splitting in the same way. Not quite sure though as I haven't debugged the code to this extend previously...

andrewdalpino:
Cool @kroky 

Did you already consider forcing there to be samples in both right and left subsets during the splitting phase?

Seems to me we should terminate the entire branch (most efficient thing to do for search) as soon as we realize it contains all the same samples instead of recursively splitting the same samples. Would a Hypersphere radius of 0.0 tell us the node contains all the same samples? I love how this bug fix is also kinda an optimization :)

kroky:
Yes, sounds good to me. However, left is always empty if the set contains all the same samples. I am not sure about Hypersphere radius - haven't yet wrapped my head around that Math but would love to one day!

andrewdalpino:
> haven't yet wrapped my head around that Math

A hypersphere is defined by a centroid (vector) and a radius (scalar). If a Hypersphere node (Ball or Clique node) contains all the same samples, then its centroid will be the same as the samples. Thus, the distance between each sample and the centroid will be 0. Radius is defined as the distance between the centroid and it's farthest neighbor. Since the max of a set of 0's is 0 then the radius of the hypersphere would be zero. So I believe we can check the value of radius to determine if the node contains all the same samples as well. I can't think of a pathological case yet. Whether or not this approach is better/worse I don't know but it's maybe something to consider.

https://github.com/RubixML/RubixML/blob/master/src/Graph/Nodes/Ball.php#L72

Anyways, I'm out for tonight, I'll review the code tomorrow. Again, nice work @kroky I'm sure this one sent you on a goose chase haha

andrewdalpino:
Thought of another thing just now, would ...

```php
$dataset->deduplicate();
```

have prevented the infinite loop as well?

src/Graph/Trees/BallTree.php
```diff
@@ -150,9 +150,14 @@ public function grow(Labeled $dataset) : void
             if ($right->numRows() > $this->maxLeafSize) {
                 $node = Ball::split($right, $this->kernel);
 
-                $current->attachRight($node);
+                [$nleft, $nright] = $node->groups();
 
-                $stack[] = $node;
+                if ($left->empty() && $nleft->empty() && $right == $nright) {
```
andrewdalpino, line 155: Would ...

```php
if ($node->radius() === 0.0) {
```

work here instead?

tests/Graph/Trees/BallTreeTest.php
```diff
@@ -83,6 +85,32 @@ public function growNeighborsRange() : void
         $this->assertCount(1, array_unique($labels));
     }
 
+    /**
+     * @test
+     */
+    public function growWithRepetitions() : void
+    {
+        $size = 25;
```
andrewdalpino, line 93: If this variable does not change can we make it a class constant? And perhaps a little more descriptive of the data it holds such as DATASET_SIZE or NUM_SAMPLES?

tests/Graph/Trees/BallTreeTest.php
```diff
@@ -83,6 +85,32 @@ public function growNeighborsRange() : void
...
+        $sample = $dataset->sample(0);
+
+        [$samples, $labels, $distances] = $this->tree->nearest($sample, 5);
+
+        $this->assertCount(5, $samples);
+        $this->assertCount(5, $labels);
+        $this->assertCount(5, $distances);
+
+        $this->assertCount(1, array_unique($samples, SORT_REGULAR));
+        $this->assertCount(1, array_unique($labels));
```
andrewdalpino, line 111: Would ...

```php
$this->assertCount(1, array_unique($distances));
```

also pass?

andrewdalpino (commented):
I'm curious if we can optimize this by only making one method call to `radius()` instead of two to `empty()` and the check for loose equality of the datasets (correct me if I'm wrong but this requires a check of every value stored in the two datasets for equality since if cannot compare object hashes?). Also some minor comments on the test.

andrewdalpino:
Hey @kroky turns out your intel lead to the discovery of another similar bug in KD Tree ... I went ahead and fixed it in this commit https://github.com/RubixML/RubixML/commit/5c531f3142ffa1b93fb0c13f39a52b51cb79dbf2. I also tested out using the radius method and it worked nicely and is slightly faster and less verbose so I went with that. What we really needed to know was if either the hypersphere or hypercube could also be defined as a single point. If it could reduce to a point then that means the node contains all of the same sample. The way we check a hypersphere is by seeing if the radius is 0, and the way we check a hypercube is to see if it's min and max vectors are equal.

src/Graph/Trees/BallTree.php
kroky: Yes, I think your logic about hypersphere radius is correct, so this seems a nice improvement of the check.

tests/Graph/Trees/BallTreeTest.php
kroky: Definitely.

tests/Graph/Trees/BallTreeTest.php
kroky: Yes, I saw you did this - looks good to me, I just didn't want to intervene too much with other tests and especially generators - I thought blob generator does not accept stdDev param equal to zero and that's why I had to generate the repetitions myself. However, your solution with using existing generator is much better, I agree.

kroky:
Great improvements, indeed the math behind this is beautiful, in my opinion :) I saw what you mean by the radius check and shrinking down the hypersphere or hypercube to a point, so thanks for the explanations and of course for the quick merge! I hope to be able to push RubixML working models in Tiki this week...

Regarding dataset->deduplicate, I think that would have worked too but would have lost information - equal samples' labels would have been reduced to a single one, so MLT search for something would have resulted in just one of the samples rather than many of them (as they all were the same).

PR #120, opened 2020-09-28
[json] Add Rubix\ML\Other\Helpers\JSON

simplechris:
##### SUMMARY:

- Add `Rubix\ML\Other\Helpers\JSON` and throw more detailed exception messages if malformed JSON is encountered. 

Small enhancement reconstructed from some local stashes that I had laying around. Let me know if you think it's worth merging 🤷‍♂️

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
     ],
     "require": {
         "php": ">=7.2",
+        "ext-json": "*",
         "amphp/parallel": "^1.3",
         "psr/log": "^1.1",
         "rubix/tensor": "^2.0",
--- a/src/Other/Helpers/JSON.php
+++ b/src/Other/Helpers/JSON.php
@@ -1,0 +1,98 @@
+<?php
+
+namespace Rubix\ML\Other\Helpers;
+
+use RuntimeException;
+
+use const JSON_ERROR_NONE;
+use const JSON_ERROR_SYNTAX;
+use const JSON_ERROR_DEPTH;
+use const JSON_ERROR_STATE_MISMATCH;
+use const JSON_ERROR_CTRL_CHAR;
+use const JSON_ERROR_UTF8;
+use const JSON_ERROR_UTF16;
+use const JSON_ERROR_RECURSION;
+use const JSON_ERROR_UNSUPPORTED_TYPE;
+use const JSON_ERROR_INVALID_PROPERTY_NAME;
+use const JSON_ERROR_INF_OR_NAN;
+
+/**
+ * JSON
+ *
+ * A helper class providing functions relating to json.
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Chris Simpson
+ */
+class JSON
+{
+    protected const DEFAULT_DEPTH = 512;
+
+    protected const DEFAULT_OPTIONS = 0;
+
+    /**
+     * Deserialize JSON.
+     *
+     * @param string $json
+     * @param int $options
+     * @param int $depth
+     *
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -34,6 +34,7 @@
     ],
     "require": {
         "php": ">=7.2",
+        "ext-json": "*",
```
andrewdalpino, line 37: Is this line necessary if we require PHP 7.2+?

https://www.php.net/manual/en/json.installation.php#json.installation

src/Other/Helpers/JSON.php
```diff
@@ -0,0 +1,83 @@
...
+                return $value;
+            case JSON_ERROR_SYNTAX:
+                throw new RuntimeException('Malformed JSON Provided. Syntax Error.');
+            case JSON_ERROR_DEPTH:
+                throw new RuntimeException('Maximum stack depth exceeded.');
+            case JSON_ERROR_STATE_MISMATCH:
+                throw new RuntimeException('Underflow or the modes mismatch.');
+            case JSON_ERROR_CTRL_CHAR:
+                throw new RuntimeException('Unexpected control character found.');
+            case JSON_ERROR_UTF8:
```
andrewdalpino, line 63: Some of the possible errors are not listed here such as NAN and INF. Should we catch those errors or did you intend for the default case to handle those?

https://www.php.net/manual/en/function.json-last-error.php#refsect1-function.json-last-error-returnvalues

src/Other/Helpers/JSON.php
```diff
@@ -0,0 +1,83 @@
...
+     * @param int $depth
+     *
+     * @throws \RuntimeException
+     *
+     * @return string
+     */
+    public static function encode($value, int $options = self::DEFAULT_OPTIONS, int $depth = self::DEFAULT_DEPTH) : string
+    {
+        $data = json_encode($value, $options, $depth);
+        if ($data === false) {
```
andrewdalpino, line 92: Nice explicit check for false here rather than falsey

andrewdalpino (commented):
I think this will be quite useful ... I had a couple of questions about some particular JSON errors that we aren't catching in the switch statement. Other than that looks good to me!

composer.json
simplechris: Yeah, I was surprised by this! I've never _ever_ used an installation of php without json enabled. 

I added this to satisfy my IDE that complains that `'ext-json' is not listed in composer.json` 

It seems that although ext-json is enabled by default you can still compile php without it with using the `--disable-json` flag, so to use json you should technically announce it as a project dependency. 🤷‍♂️  


---
**EDIT:** This flag is removed in PHP8 meaning json will always be available (https://externals.io/message/109783)

src/Other/Helpers/JSON.php
simplechris: 👍 Yeah, let me add those in now

simplechris (commented):
@andrewdalpino updated and included all the json error states: ([diff](https://github.com/RubixML/RubixML/compare/84736969692becafb1dbf52163a0d1c93f66355c..20573cf1f6d1e5af8f59e88e31b798060f810a4d))

If you want me to remove the `ext-json` thing then totally happy to, it seems more like a theoretical edge-case and I only became aware of it via IDE notifications.

composer.json
andrewdalpino: For shizzle, thanks for checking that out

andrewdalpino (approved):

andrewdalpino:
Thanks again @simplechris, very helpful!

PR #125, opened 2020-10-05
[RFC] Data Persistence in Rubix ML

simplechris:
<table width="100%">
  <tr><th>✏️</th><td><b>[RFC] Data Persistence in Rubix ML</b></td></tr>
  <tr><th>👨‍💻 </th><td> <a href="https://github.com/simplechris">Chris Simpson</a> </td></tr>
  <tr><th>🗓 </th><td>September - October 2020</td></tr>
</table>

---

#### BACKGROUND:

It is very common for a user of the RubixML library to load or save data. 

Within RubixML most actions relating to [data persistence](https://en.wikipedia.org/wiki/Persistence_(computer_science)) are typically achieved via the use of [`Persister`](https://docs.rubixml.com/en/latest/persisters/api.html#persisters) either directly, or via the [`PersistentModel`](https://docs.rubixml.com/en/latest/persistent-model.html) class. The purpose of a [`Persister`](https://docs.rubixml.com/en/latest/persisters/api.html#persisters) is to _save_ a new [`Persistable`](https://docs.rubixml.com/en/latest/persistable.html) object or to _load_ a previously saved object.

Whilst a [`Persister`](https://docs.rubixml.com/en/latest/persisters/api.html#persisters) handles the persistence of objects (recursively [de]serializing the objects and their internal dependencies), the more-general action of _reading_ or _writing_ data also appears in numerous other places within the library e.g. The [`Encoding`](https://github.com/RubixML/RubixML/blob/master/src/Encoding.php) object has a [`write()`](https://github.com/RubixML/RubixML/blob/master/src/Encoding.php#L51) method used to _write_ data to the local filesystem, whilst the purpose of [`Extractors`](https://docs.rubixml.com/en/latest/extractors/api.html#extractors) is to _read_ data and transform it into [`Dataset`](https://docs.rubixml.com/en/latest/datasets/api.html#dataset-objects) objects.

**This contribution aims to introduce abstractions for data IO suitable for use across the entire codebase. It standardizes the mechanism of how data is persisted (saved/loaded) within RubixML.**

#### DESIGN GOALS:

- Remove any tight-coupling to the local file system: Introduce a new abstraction modelling where data can be stored (local filesystem, remote/cloud storage, database etc). The default location can remain as the local filesystem, but alternative implementations should be used if provided.
- Split the responsibility of _reading_ and _writing_ data. By design [`Extractors`](https://docs.rubixml.com/en/latest/extractors/api.html#extractors) should be read-only, whilst [`Encoding`](https://github.com/RubixML/RubixML/blob/master/src/Encoding.php) objects need only have awareness of how to _write_ the data that they contain.
- Retain the simplicity and of the [`Persister`](https://docs.rubixml.com/en/latest/persisters/api.html) interface, whilst introducing lower-level abstractions to power _all_ current and future data storage operations.
- Allow for the incremental and memory-efficient reading and writing of data via the use of [steams](https://www.php.net/manual/en/intro.stream.phphttps://www.php.net/manual/en/intro.stream.php) and [generators](https://www.php.net/manual/en/language.generators.overview.php) wherever possible/practical.

#### APPROACH:

This PR introduces the `Rubix\ML\Storage` namespace. This namespace contains abstractions relating to _reading_ and _writing_ of data: `Datastore` defines the behaviour of a generic data-storage repository. It is given the ability to perform _read_-related operations via the `Rubix\Ml\Storage\Reader` interface, and it's ability to perform _write_-related operations from the `Rubix\Ml\Storage\Writer` interface. A `Datastore` fully implements both of these interfaces. This separation allows for `Reader` to be type-hinted where read-only behaviour is intended. `Writer` to be used where the class only wishes to _write_ data. `Datastore` that encompasses both `Reader` and `Writer` can be hinted to classes that require both mechanisms (eg `Persister` implementations).

##### `Reader`:

```php
<?php

namespace Rubix\ML\Storage;

use Rubix\ML\Storage\Streams\Stream;

/**
 * Reader
 *
 * @category    Machine Learning
 * @package     Rubix/ML
 * @author      Chris Simpson
 */
interface Reader
{
    /**
     * Return if the target exists at $location.
     *
     * @param string $location
     * @throws \Rubix\ML\Storage\Exceptions\StorageException
     * @return bool
     */
    public function exists(string $location) : bool;

    /**
     * Open a stream of the target data at $location.
     *
     * @param string $location
     * @param string $mode
     * @throws \Rubix\ML\Storage\Exceptions\ReadError
     * @throws \Rubix\ML\Storage\Exceptions\StorageException
     * @return \Rubix\ML\Storage\Streams\Stream
     */
    public function read(string $location, string $mode = Stream::READ_ONLY) : Stream;
}

```

##### `Writer`:

```php
<?php

namespace Rubix\ML\Storage;

/**
 * Writer
 *
 * @category    Machine Learning
 * @package     Rubix/ML
 * @author      Chris Simpson
 */
interface Writer
{
    /**
     * Write.
     *
     * @param string $location
     * @param mixed $data
     * @throws \Rubix\ML\Storage\Exceptions\WriteError
     * @throws \Rubix\ML\Storage\Exceptions\StorageException
     */
    public function write(string $location, $data) : void;

    /**
     * Move.
     *
     * NOTE: If supported by the underlying datastore this should be implemented as an atomic operation.
     *
     * @param string $from
     * @param string $to
     * @throws \Rubix\ML\Storage\Exceptions\StorageException
     */
    public function move(string $from, string $to) : void;

    /**
     * Delete.
     *
     * @param string $location
     * @throws \Rubix\ML\Storage\Exceptions\StorageException
     */
    public function delete(string $location) : void;
}

```

##### `Datastore`:

```php
<?php

namespace Rubix\ML\Storage;

use Stringable;

/**
 * Datastore.
 *
 * Defines the behaviour of a generic storage repository (filesystem, database etc)
 *
 * @category    Machine Learning
 * @package     Rubix/ML
 * @author      Chris Simpson
 */
interface Datastore extends Reader, Writer, Stringable
{
    //
}

```

##### `Stream`:

`Reader::read()` returns an object implementing the `Stream` interface. This interface acts as an [OO](https://en.wikipedia.org/wiki/Object-oriented_programming) wrapper around php [streams](https://www.php.net/manual/en/intro.stream.php). This functionality is explicitly wrapped to enable typehinting (`resource` is not a valid language-level typehint), and for consistent non repetitive stream interaction. `Stream` represents a generic stream of data and implements all common stream operations via methods named after the common semantics used.

The `Stream` interface enables the ability to incrementally (in a 'cursorable' manner) read the data from the underlying resource, and also implements [`IteratorAggregate`](https://www.php.net/manual/en/class.iteratoraggregate.php) allowing the object to be iterated upon directly (yielding a line of content per iteration) Note: To read the entire contents of the stream in a single operation see `Stream::contents()`.

#### INTEGRATION:

To illustrate the implementation and integration of these new objects I have provided an example: `Rubix\ML\Storage\LocalFilesystem`. This is the simplest of implementations and can be used to replace any existing interaction with data stored on the local filesystem for example `Extractor` implementations, `Encoding` objects, and the widely used `Filesystem` Persistor class.

##### `Rubix\ML\Storage\LocalFilesystem`:

```php
<?php

namespace Rubix\ML\Storage;

use Rubix\ML\Storage\Streams\File;
use Rubix\ML\Storage\Streams\Stream;
use Rubix\ML\Storage\Exceptions\RuntimeException;

/**
 * Local Datastore.
 *
 * @category    Machine Learning
 * @package     Rubix/ML
 * @author      Chris Simpson
 */
class LocalFilesystem implements Datastore
{
    /**
     * @see \Rubix\ML\Storage\Reader::exists()
     * @inheritdoc
     *
     * @param string $location
     *
     * @return bool
     */
    public function exists(string $location) : bool
    {
        return file_exists($location);
    }

    /**
     * @see \Rubix\ML\Storage\Reader::read()
     * @inheritdoc
     *
     * @param string $location
     *
     * @return \Rubix\ML\Storage\Streams\Stream
     */
    public function read(string $location, string $mode = Stream::READ_ONLY) : Stream
    {
        $resource = fopen($location, $mode);

        if (!$resource) {
            throw new RuntimeException("Could not open $location.");
        }

        $stream = new File($resource);

        if (!$stream->readable()) {
            throw new RuntimeException("Stream with mode {$mode} cannot be read from");
        }

        return $stream;
    }

    /**
     * @see \Rubix\ML\Storage\Writer::write()
     * @inheritdoc
     *
     * @param string $location
     * @param \Rubix\ML\Storage\Streams\Stream|string $data
     */
    public function write(string $location, $data) : void
    {
        if ($data instanceof Stream) {
            $data = $data->contents();
        }

        file_put_contents($location, $data, LOCK_EX);
    }

    /**
     * @see \Rubix\ML\Storage\Writer::delete()
     * @inheritdoc
     *
     * @param string $location
     */
    public function delete(string $location) : void
    {
        unlink($location);
    }

    /**
     * @see \Rubix\ML\Storage\Writer::move()
     * @inheritdoc
     *
     * @param string $from
     * @param string $to
     */
    public function move(string $from, string $to) : void
    {
        rename($from, $to);
    }

    /**
     * Return the string representation of the object.
     *
     * @return string
     */
    public function __toString() : string
    {
        return 'Local Filesystem';
    }
}

```

##### `Encoding`:

The `Encoding::write()` method now accepts an optional second parameter of type `Rubix\ML\Storage\Writer`. Omitting this argument will cause an instance of `Rubix\ML\Storage\LocalFilesystem` to be used (so existing behaviour is transparently maintained resulting in no BC breaks)

##### `Extractors`:

By design, `Extractor` objects are [read-only](https://en.wikipedia.org/wiki/File-system_permissions). All existing implementations now accept a `Reader` in their constructors and the implementations have been updated to show how the `Stream` object returned by `Reader::read(): Stream` can be used in-place of the existing semantics. 

*Note:* Any `Reader` implementation can be used: so `Extractors` are now decoupled from the local filesystem and able to read data from a multitude of storage backends.

---

#### DISCUSSION:

Whilst I have given this implementation considerable thought, and gone though numerous iterations prior to opening this PR, I still think there is room for a significant discussion here. The specific implementations here are intended as illustrative and I'm most interested in feedback on the design itself. I have a number of open questions myself, but think that this is mature enough to introduce external feedback and inspection.

- **Naming is hard.** Whilst I've tried to be as concise as possible I think there is room to improve the naming of these abstractions! Which is preferable: `Datastore`? `Repository`? `StorageOperator`, `StorageEngine` etc? Happy to take suggestions! I would have initially used `Backend` but obviously didn't want to cause any confusion with the [`Backend`](https://docs.rubixml.com/en/latest/parallel.html#set-a-backend) interface used in the parallel processing.

- **Future proof?** I've tried to make these abstractions generic and multi-purpose. If you can anticipate a future use-case that wouldn't be supported then let's discuss.

- **Backwards Compatibility**: Have I broken anything? Should I have? Let's talk about the options here.

- **Unit Tests**: Whilst all current unit tests are passing, I have purposefully neglected to add further tests prior to confirming the design/discussion/direction in this PR.

- **Exceptions**: The exceptions were added prior to the merge of https://github.com/RubixML/RubixML/pull/116 so are more a rudimentary demonstration of potential flexibility and not a active recommendation of how we should structure an exception hierarchy.

- **Pull Request**: At the time of writing I have based this PR against `0.3.0`, as it contains the most up-to-date 'nightly' code. This is just to minimise the size of the diff.

---
#### RELATED:
- https://github.com/RubixML/RubixML/issues/108: How can the persistence subsystem be extended across the codebase? This was my starting point for this PR. I realized that Persisters are actually very good at what they do (serializing objects and storing them) and backing them by the common storage abstractions (without changing the `Persister` interface) would be a positive step forwards.
- A early proof-of-concept showing how storage could be made pluggable using `Flysystem 1.x`: https://github.com/RubixML/RubixML/pull/106.
- A Persistor backed by `Flysystem 2.x` (beta): https://github.com/RubixML/Extras/pull/3

###### `2020-10-05`:
- Initial publish and request for comments.

### DIFF
--- a/src/Encoding.php
+++ b/src/Encoding.php
@@ -3,6 +3,10 @@
 namespace Rubix\ML;
 
 use Rubix\ML\Exceptions\RuntimeException;
+use Rubix\ML\Storage\Exceptions\StorageException;
+use Rubix\ML\Storage\LocalFilesystem;
+use Rubix\ML\Storage\WriteProxy;
+use Rubix\ML\Storage\Writer;
 use Stringable;
 
 class Encoding implements Stringable
@@ -46,22 +50,21 @@ class Encoding implements Stringable
      * Write the encoding to a file at the path specified.
      *
      * @param string $path
+     * @param ?Writer $storage
      * @throws \Rubix\ML\Exceptions\RuntimeException
      */
-    public function write(string $path) : void
+    public function write(string $path, ?Writer $storage = null) : void
     {
-        if (!is_file($path) and !is_writable(dirname($path))) {
-            throw new RuntimeException('Folder does not exist or is not writable');
+        if (!$storage) {
+            $storage = new LocalFilesystem();
         }
 
-        if (is_file($path) and !is_writable($path)) {
-            throw new RuntimeException("File $path is not writable.");
-        }
+        $storage = new WriteProxy($storage);
 
-        $success = file_put_contents($path, $this->data, LOCK_EX);
-
-        if (!$success) {
-            throw new RuntimeException('Failed to write to the filesystem.');
+        try {
+            $storage->write($path, $this->data);
+        } catch (StorageException $e) {
+            throw new RuntimeException($e->getMessage(), $e->getCode(), $e);
         }
     }
 
--- a/src/Extractors/CSV.php
+++ b/src/Extractors/CSV.php
@@ -4,6 +4,11 @@ namespace Rubix\ML\Extractors;
 
 use Rubix\ML\Exceptions\InvalidArgumentException;
 use Rubix\ML\Exceptions\RuntimeException;
+use Rubix\ML\Storage\Exceptions\StorageException;
+use Rubix\ML\Storage\LocalFilesystem;
+use Rubix\ML\Storage\Reader;
+use Rubix\ML\Storage\ReadProxy;
+use Rubix\ML\Storage\Streams\Stream;
 use Generator;
 
 use function strlen;
@@ -29,11 +34,11 @@ use function strlen;
 class CSV implements Extractor
 {
     /**
-     * The file handle.
+     * A readable stream object pointing to the underlying CSV data
      *
-     * @var resource
+     * @var \Rubix\ML\Storage\Streams\Stream
      */
-    protected $handle;
+    protected $stream;
 
     /**
      * Does the CSV document have a header as the first row?
@@ -57,25 +62,29 @@ class CSV implements Extractor
     protected $enclosure;
 
     /**
-     * @param string $path
+     * @param string $location
      * @param bool $header
      * @param string $delimiter
      * @param string $enclosure
-     * @throws \Rubix\ML\Exceptions\InvalidArgumentException
-     * @throws \Rubix\ML\Exceptions\RuntimeException
+     * @param ?Reader $storage
+     * @throws \InvalidArgumentException
+     * @throws \Rubix\ML\Storage\Exceptions\StorageException
      */
     public function __construct(
-        string $path,
+        string $location,
         bool $header = false,
         string $delimiter = ',',
-        string $enclosure = '"'
+        string $enclosure = '"',
+        ?Reader $storage = null
     ) {
-        if (!is_file($path)) {
-            throw new InvalidArgumentException("Path $path does not exist.");
+        if (!$storage) {
+            $storage = new LocalFilesystem();
         }
 
-        if (!is_readable($path)) {
-            throw new InvalidArgumentException("Path $path is not readable.");
+        $storage = new ReadProxy($storage);
+
+        if (!$storage->exists($location)) {
+            throw new InvalidArgumentException("Location $location does not exist.");
         }
 
         if (strlen($delimiter) !== 1) {
@@ -88,56 +97,59 @@ class CSV implements Extractor
                 . ' a single character, ' . strlen($enclosure) . ' given.');
         }
 
-        $handle = fopen($path, 'r');
-
-        if (!$handle) {
-            throw new RuntimeException("Could not open $path.");
-        }
-
-        $this->handle = $handle;
+        $this->stream = $storage->read($location, Stream::READ_ONLY);
         $this->header = $header;
         $this->delimiter = $delimiter;
         $this->enclosure = $enclosure;
     }
 
     /**
-     * Clean up the file pointer.
+     * Clean up the any open file handles.
      */
     public function __destruct()
     {
-        fclose($this->handle);
+        try {
+            if ($this->stream and $this->stream->open()) {
+                $this->stream->close();
+            }
+        } catch (StorageException $e) {
+            //
+        }
     }
 
     /**
      * Return an iterator for the records in the data table.
      *
      * @throws \Rubix\ML\Exceptions\RuntimeException
+     * @throws \Rubix\ML\Storage\Exceptions\StorageException
      * @return \Generator<mixed[]>
      */
     public function getIterator() : Generator
     {
-        rewind($this->handle);
+        if ($this->stream->seekable()) {
+            $this->stream->rewind();
+        }
 
         $line = 0;
 
         if ($this->header) {
-            $header = fgetcsv($this->handle, 0, $this->delimiter, $this->enclosure);
+            $row = $this->stream->line();
 
-            if (!$header) {
+            if (empty(trim($row))) {
                 throw new RuntimeException('Header not found on the first line.');
             }
 
+            $header = str_getcsv($row, $this->delimiter, $this->enclosure);
             ++$line;
         }
 
-        while (!feof($this->handle)) {
-            $record = fgetcsv($this->handle, 0, $this->delimiter, $this->enclosure);
-
-            ++$line;
-
-            if (empty($record)) {
+        foreach ($this->stream as $row) {
+            if (empty($row)) {
                 continue 1;
             }
+
+            $record = str_getcsv($row, $this->delimiter, $this->enclosure);
+            ++$line;
 
             if (isset($header)) {
                 $record = array_combine($header, $record);
--- a/src/Extractors/JSON.php
+++ b/src/Extractors/JSON.php
@@ -4,8 +4,13 @@ namespace Rubix\ML\Extractors;
 
 use Rubix\ML\Exceptions\InvalidArgumentException;
 use Rubix\ML\Exceptions\RuntimeException;
-use Generator;
 use Rubix\ML\Other\Helpers\JSON as JSONHelper;
+use Rubix\ML\Storage\Exceptions\StorageException;
+use Rubix\ML\Storage\LocalFilesystem;
+use Rubix\ML\Storage\Reader;
+use Rubix\ML\Storage\ReadProxy;
+use Rubix\ML\Storage\Streams\Stream;
+use Generator;
 
 /**
  * JSON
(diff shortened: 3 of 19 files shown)


simplechris:

simplechris:
recheckcla

simplechris:
The initial submission here deliberately includes no BC breaks, but does manage to make all places that save or load data use common abstractions, and doesn't couple us to any external library or interface. Through discussion I would like to fine-tune these interfaces and implementations. 

The particularly interesting improvements come from even deeper integration, or considering breaking API changes too. I have built on the above submission locally: a lot of flexibility is unlocked by making the `Stream` interface `Persistable` or even better: `Encoding` implement `Stream`. 

I didn't want to introduce too much in one go without discussing it but there are some cool directions to take now.

No rush, though!

marclaporte:
BC breaks: Seems to me it's quite OK to do now given Rubix ML is quite young. Just follow semver so folks know what to expect.

marclaporte:
Victor checked it and he shared "Looks like a very well thought PR and needed in my opinion. Couldn't find something to contribute, I'd simply use it when it is out :-)"

andrewdalpino:
Beautiful PR/RFC @simplechris I finally got around to reading the whole thing and 1. Thank you for such a well thought description and 2. thank you for all the time it took in architecting this solution. I think this is probably the way forward. Would you call this a 'repository' pattern (i.e. are DataStore and Repository functionally equivalent)? I like the name DataStore and the names Reader and Writer make perfect sense. I like the interfaces. I need to take some more time to understand the Streams stuff and why we need a WriteProxy (we can't use a Writer instance directly instead?) but I'm liking the DataStore abstraction.

Future Proof? It's hard to tell lol! I don't know to be honest but it's much better than what we have so it's *more* futureproof in that regard :)

I don't see an BC breaks just yet, I plan to do a more in-depth review of the code I just wanted to get a convo going while I had some free time. Overall very impressed. Nice work!

andrewdalpino:
recheckcla

simplechris:
Thanks for the comments @marclaporte, @andrewdalpino 👍 

---

- Regarding `Repository`: I was thinking along the lines of the dictionary definition: _a place where or receptacle in which things are or may be stored_ but agree that this word already has existing meaning within software design and could easily be a source of confusion to users. I'm good with `DataStore` if you are 👍 

- Regarding `ReadProxy` etc: Yep, I was expecting this to be called out. This was intended to decorate a `Datastore` (which has `Reader` AND `Writer` functionality and act as a protection to ensure that nobody could unwittingly or erroneously use other behaviour that the object has when in a read-only or a write-only context. In reality: static analysis will enforce this as calling `read()` on a `Writer` wouldn't be allowed (even though the datastore object contains further behaviour, the `Reader` or `Writer` typehint would enforce only a subset of the functionality could be used). Fine with removing these.

- Regarding the `Stream` abstraction - I think its quite neat. A (generic) datastore can read data and represent it the form of a stream. Datastore implementations incrementally read these from a socket, or just buffer these through php temp streams. It allows for consistent 'cursorable'/line-by-line processing, `$datastore->read('location_of_data')->line()` (Or just iterating over the stream directly) but can equally simply return the entire stream as a string in the situations where that is desired. `$datastore->read('location_of_data')->contents()`. I think it's flexible but also fairly clean too. Could use some further refinement but i feel that it could be a helpful abstraction to have around..

andrewdalpino:
Regardless of semantics @simplechris, is this pattern similar to the one outlined here?

https://martinfowler.com/eaaCatalog/repository.html

azuso:
@andrewdalpino not completely, but holds some similarities. 

It’s not a ‘repository’ in exactly the same sense as you would have in data-mappers/ORMs - like a Doctrine repository for PHP - but merely defines abstract data-access operations. What do you think?

simplechris:
> Regardless of semantics @simplechris, is this pattern similar to the one outlined here?
> 
> 
> 
> https://martinfowler.com/eaaCatalog/repository.html

@andrewdalpino not completely, but holds some similarities. 

It’s not a ‘repository’ in exactly the same sense as you would have in data-mappers/ORMs - like a Doctrine repository for PHP - but merely defines abstract data-access operations. What do you think?

andrewdalpino:
The first thing I thought of was repository when I saw the Data Store interface but it doesn't really matter

I'm hoping to give this another look soon focusing on the Stream stuff, it sounds creative

andrewdalpino:
Can you retarget master with this @simplechris?

andrewdalpino:
Hey @simplechris how are you feeling about this design now that we've had some time to let it sit in?

andrewdalpino:
Yo @simplechris I circled back on this once more and I think I'm at the point where I feel this is too much complexity for what we need. I'm actually starting to think that we had it right in the first place (ununified persistence). It seems there are fundamental differences in persisting an object, reading/writing a dataset, and saving arbitrary text. I think you did a great job of capturing what we need, but the result is that we have to create so many abstractions just to get everything wired up together. I think the best way forward is to *embrace* the unique differences between the things we persist and to build on the custom abstractions that cater to their individual needs such as we currently do with the Persister/Serializer subsystem designed for persisting PHP objects. For example, one thing we could do is put a method on the Extractor API that takes a dataset and writes it to disk in its own format. In this case, each implementation can take care of streaming if compatible. Since we are currently developing 1.0, now would be a good time to have another discussion. Let me know if you are down!

andrewdalpino:
Going one deeper ...

Persisters - Save and load objects in efficient non-human readable blobs

Extractors - Save and load data tables stored line by line

Encoding - Save and load arbitrary strings (may need to be human readable) as blobs

I think we have good abstractions around Persisters and Extractors now in 1.0 but I think we can still work on a more flexible way to persist arbitrary strings via the Encoding value object. We could unify "all things persisted as a blob" using the Persister interface but we'd have to flip the Serializer to the outside of the abstraction.

Ex. Persistable -> Serializer -> Encoding -> Persister ... for Persistable objects

Ex. Report Generator -> Report -> Encoding -> Persister ... for Report objects

The problem with this is that we lose some sugar. The loss in sugar can be made up by decorating an estimator or Pipeline with Persistent Model, however there is no obvious way to make persisting individual transformers nice and clean. Having that said, I'm sure we can think of something if we put our minds to it!

PR #126, opened 2020-10-06
Variable length of Skip-Gram tokens

absolutic:


absolutic:

absolutic:
recheckcla

simplechris:
hey @absolutic, welcome 👋 

Checks are not currently passing as the commit doesn't conform to the coding standard. `composer fix` will automatically format the code to be consistent with the coding standard 👍

absolutic:
_**I have read the CLA Document and I hereby sign the CLA**_

absolutic:
recheckcla

absolutic:
Hello @simplechris, did it, thanks! The build seems to be fine now.

andrewdalpino:
Hey @absolutic It looks like CLA bot changed the name of some configuration options and it broke our setup (they are in Alpha still so that can sometimes happen). I updated the config ... can you try pulling those changes into your branch and then push again after merge so this PR has the new correct options?

andrewdalpino (approved):
Clean and elegant solution @absolutic! Great execution, thank you for the PR and welcome to our community!

Since replacing the current implementation of Skip Gram will cause a BC break we'd have to hold off until 1.0.0 (not until 2021 for sure) to release it - but I think I have a better idea. What if we called your implementation something like "KSkipNGram" and introduce it as a feature in  0.3.0 (this or next month) while we deprecate SkipGram?

If you're down just rename the class and add a documentation page and link to it in https://github.com/RubixML/RubixML/blob/master/mkdocs.yml. Also, you'll need to change the target to the 0.3.0 branch and use our new Exception classes (Rubix\ML\Exceptions\InvalidArgumentException) instead of the SPL ones (\InvalidArgumentException). Lastly, instead of the sentence regex, you can use the new Sentence tokenizer under the hood (see the 0.3.0 branch NGram or SkipGram for an example).

Also feel free to join us in our Telegram channel https://t.me/RubixML

absolutic:
Hello @andrewdalpino, thank you for your feedback about my PR. Yes, sure, I will refactor it tomorrow according to your idea with KSkipNGram and following all recommendations. I will push bot config as a part of my following changes.

andrewdalpino:
Per our discussion, this will be implemented as a separate tokenizer to avoid BC breaks

PR #158, opened 2021-01-22
RBX file format

andrewdalpino:
This is an implementation born out of the research here https://github.com/RubixML/ML/issues/153 to solve this problem https://github.com/RubixML/ML/issues/115 among others.

This idea is fairly simple, we automatically track class revisions and use a custom file format with meta data to keep track of the version. By default, the RBX format uses Gzipped native PHP serialization. An optional signing password can be given in the constructor to prevent tampering. Under the hood, the serializer employs a two-stage HMAC verification using the supplied password to simultaneously check the integrity as well as authenticity of both the header and payload. This operation is almost always fast for invalid payloads as only the (smaller) header need be hashed to determine tampering in 99% of the cases. The rest who've managed to forge the header will (hopefully) get caught by the second (slower) stage that verifies the payload data.

All in all, I'm pretty happy with the value we're able to add by building on top of well-known solutions and I'm hoping this format will give more users even more confidence to both share their models and to store them on their servers.


src/Other/Traits/TracksRevisions.php
```diff
@@ -0,0 +1,32 @@
...
+trait TracksRevisions
+{
+    /**
+     * The revision number of the class.
+     *
+     * @return string
+     */
+    public function revision() : string
+    {
+        $properties = array_keys(get_object_vars($this));
```
ChristophWurst: what if the type changes? This will be more relevant with typed props in php7.4

ChristophWurst:
This makes sense, though I wonder if this is an unsolved problem in the php ecosystem and there aren't any sophisticated solutions (packages) out there that you could base your serialization on.

> All contributors have signed the CLA

On a side node I'm surprised to see a CLA on this project. Is there any info on why this is required?

src/Other/Traits/TracksRevisions.php
andrewdalpino: This method is not perfect, there's other limitations as well that I'm aware of

This is more of a POC for now ... we could hash the property names *and* types when we require PHP 8.0+ for sure

andrewdalpino:
> On a side node I'm surprised to see a CLA on this project. Is there any info on why this is required?

Primarily to aid in litigation if that (hopefully never) happens (again). It means that the current maintainers have the rights to represent all contributors. All sorts of other stuff focusing on longevity but the legal reasons were the primary motivator.

https://github.com/RubixML/ML/blob/master/CONTRIBUTING.md

https://github.com/RubixML/Signatures/blob/master/CLA.md

Does NextCloud have a CLA too?

andrewdalpino:
Thanks for the review @ChristophWurst!

> This makes sense, though I wonder if this is an unsolved problem in the php ecosystem and there aren't any sophisticated solutions (packages) out there that you could base your serialization on.

I am not aware of any, are you?

While we're at it, are there any other features you'd like to see related to this?

andrewdalpino:
So we came up with two candidate file formats ... RBX and RBXE (Encrypted). We might go with one of them, or both of them. I'm interested in hearing some feedback.

Both formats are based on the native PHP serialization format, have a secure header via sha256 HMAC, and add checks such as payload length and class revision number.  Where they differ is the following ...

RBX standard does not encrypt the payload, rather it signs it with another sha256 HMAC. Thus it does not require the open_ssl extension. The downside is that the data can still be read by someone snooping. The upside is that the data can be compressed without weakening security.

RBXE encrypts the payload with aes256 and therefore requires the open_ssl extension. This solves the problem of snooping but the data cannot be compressed (see https://en.wikipedia.org/wiki/CRIME).

As far as file sizes go, they are both about the same. Both formats allow you to omit a password by default and password digests are resilient to brute-force attacks. In addition, hash comparisons are evasive to timing attacks. When passwords are omitted, the HMACs and encryption effectively degrade into checksums.

3 scenarios in which these features become very useful ...

1. You run a model on a server and want to ensure that it's tamper-proof
2. You want to share a model over a public network and ensure it's integrity
3. Your model is part of an automated system and you want to know when library upgrades should trigger a retrain

Here's what the files generally look like under the hood ...

```
�RBXE

sha256:1cb9a753e3b26ef4648642e53f02bd903c827f35e437b85ee2024eb3894be0f7
{"version":1,"class":{"name":"Rubix\\ML\\Pipeline","revision":"0e5efcc2"},"data":{"encryption":{"method":"aes256","iv":"xm9Whwk+BXTfQeGP1boHUw=="},"length":703648}}
��sk���Ab�\��m�/[k4��`X�u�sY�u"akXa4�V������5W���B�j�R��9��zu��Ώ�i1���Fk豨3�b��6�e��Vߐ�+d�om��l�*kG�U���Rٵ���U$�"�տ{@ew�M{��`@s*1����hݲ6Vp��v
G�X��Tٸ�ؑB��{;5�l�u�F�q����|�۲���x�<>
```

Any thoughts @simplechris @ChristophWurst?

I'm wondering if we should track the version number of the library so we can say something like "Ooops wrong version, downgrade to 0.2.4 to load this object etc." Maybe in a future iteration ... it's kind of a pain to track version constants.

PR #159, opened 2021-01-31
New Transformer: Boolean Converter

exzachlyvv:
The boolean converter can be used to transform `true` and `false` values in datasets to the continuous equivalent 1 and 0 respectively.


exzachlyvv:

andrewdalpino:
Interesting @exzachlyvv thanks for the PR

Summarizing our convo on Telegram .... I think we're on the right track if we can allow the user to control the representation of both true and false values. The other considerations are automatically handling booleans in  the dataset object, or not doing anything and letting the user manually choose the representation of their boolean values. I'm liking the Transformer approach as it gives the user maximum control at the slight cost of a preprocessing step.

```php
$transformer = new BooleanConverter($trueValue = 'true', $falseValue = 'false);

$transformer = new BooleanConverter($trueValue = 'tall', $falseValue = 'not tall');

$transformer = new BooleanConverter($trueValue = 1, $falseValue = 0);
```

I'm 99% sure on the transformer with the extra customization ... if you're highly confident about it as well I'd say let's go for it. Probably should limit the converted value to strings and ints (floats maybe?).

For reference

https://github.com/RubixML/ML/blob/master/src/Datasets/Labeled.php#L136

andrewdalpino (approved):

src/Transformers/BooleanConverter.php
```diff
@@ -0,0 +1,95 @@
...
+    protected $falseValue;
+
+    /**
+     * BooleanConverter constructor.
+     * @param mixed $trueValue
+     * @param mixed $falseValue
+     */
+    public function __construct($trueValue = 'true', $falseValue = 'false')
+    {
+        if (!is_float($trueValue) && !is_int($trueValue) && !is_string($trueValue)) {
```
andrewdalpino: Would `is_numeric()` work here in place of the calls to `is_float()` and `is_int()`?

src/Transformers/BooleanConverter.php
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Transform the dataset in place.
+     *
+     * @param array[] $samples
+     */
+    public function transform(array &$samples) : void
+    {
+        foreach ($samples as &$sample) {
+            foreach ($sample as &$value) {
+                if (is_bool($value)) {
```
andrewdalpino, line 80: One thing we can do to optimize this is to determine the columns that have type "Other" during a fitting process and then only iterate over those columns. Is that something you considered in your approach? If not, perhaps we could try it out to see if it will yield a justifiable performance boost.

I can help you create a benchmark for this Transformer so you we can measure the difference in approaches.

P.S. Another smaller optimization would be to import the `is_bool()` function explicitly (ex. import function is_bool;). These "is" functions are special and can be optimized by the Zend engine if it knows they aren't being overridden (hence the explicit import).

src/Transformers/BooleanConverter.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Return the string representation of the object.
+     *
+     * @return string
+     */
+    public function __toString() : string
+    {
+        return 'Boolean Converter';
```
andrewdalpino: Would it be possible to add the constructor params as part of this object's string representation? Similar to how we do here ...

https://github.com/RubixML/ML/blob/master/src/Transformers/MinMaxNormalizer.php#L214

tests/Transformers/BooleanConverterTest.php
```diff
@@ -0,0 +1,60 @@
+<?php
```
andrewdalpino, line 1: Sick test

docs/transformers/boolean-converter.md
```diff
@@ -0,0 +1,30 @@
...
+| # | Name | Default | Type | Description |
+|---|---|---|---|---|
+| 1 | trueValue | 'true' | string, int, float | The value to convert `true` to. |
+| 2 | falseValue | 'false' | string, int, float | The value to convert `false` to. |
+
+## Example
+```php
+use Rubix\ML\Transformers\BooleanConverter;
+
+$transformer = new BooleanConverter($trueValue = 'true', $falseValue = 'false);
```
andrewdalpino: Serious question, is this correct syntax for PHP 8 named arguments?

Even if it is, maybe use the old school method of injecting constructor parameters for now until PHP really catches on

andrewdalpino (commented):
Hey @exzachlyvv just a few suggestions ... I'd be curious to know if we can make this faster using information we already have computed in the dataset object during fitting. We'd basically implement the Stateful interface and inside `fit()` we'd memorize the columns that had the "Other" data type.

Let me know if you'd like to experiment ... we can set up a benchmark if so

Otherwise looks good just a few minor comments

andrewdalpino:
Hey @exzachlyvv I went ahead and ran some tests ... it's not looking like it's worth it to memorize the types during fitting ... I have a few hunches but can't say for sure why. The gains were negligible and given the extra code it's feeling like a micro optimization.

Let me know if you still wanted to run your own experiments ... Here's the benchmark I created to test Numeric String Converter

https://github.com/RubixML/ML/blob/0.4.0/benchmarks/Transformers/NumericStringConverterBench.php

If not, don't worry about it then we're good to go!

docs/transformers/boolean-converter.md
exzachlyvv: From: https://stitcher.io/blog/php-8-named-arguments

Example of named arguments in PHP 8:
```
setcookie(
    name: 'test',
    expires: time() + 60 * 60 * 2,
);
```

exzachlyvv:
@andrewdalpino I am good to merge this. The failing checks do not look related to my changes. Is there something I should do to fix those up?

andrewdalpino:
Hmm not sure why Stan is complaining but no worries goo to go!

Thank you again @exzachlyvv, nice work, elegant solution!

PR #160, opened 2021-01-31
Fix broken table markdown formatting on Choosing an Estimator

exzachlyvv:
Right now, 2 tables have broken markdown format. This PR fixed them.

image


andrewdalpino (approved):
Dude nice catch

PR #175, opened 2021-04-29
Make MaxAbsoluteScaler bidirectional (supporting reverse transformation)

torchello:


src/Datasets/Dataset.php
```diff
@@ -290,6 +291,30 @@ public function apply(Transformer $transformer) : self
...
+     * @throws \Rubix\ML\Exceptions\RuntimeException
+     * @return static
+     */
+    public function reverse(Transformer $transformer) : self
+    {
+        if (!$transformer instanceof Bidirectional) {
+            throw new RuntimeException('Transformer does not support reverse transformation.');
+        }
+
+        if ($transformer instanceof Stateful) {
```
andrewdalpino, line 303: If the transformer is Stateful and is not fitted, should it throw an exception instead? In order to reverse the transformation we'd need the information calculated at fitting no?

src/Datasets/Dataset.php
```diff
@@ -290,6 +291,30 @@ public function apply(Transformer $transformer) : self
...
     }
 
+    /**
+     * Reverse a transformation to the dataset.
+     *
+     * @param \Rubix\ML\Transformers\Transformer $transformer
+     * @throws \Rubix\ML\Exceptions\RuntimeException
+     * @return static
+     */
+    public function reverse(Transformer $transformer) : self
```
andrewdalpino: I think we can come up with a more precise name for this method. Let's keep this in mind.

EDIT: What do you think of `reverseApply()`?

src/Datasets/Dataset.php
```diff
@@ -290,6 +291,30 @@ public function apply(Transformer $transformer) : self
...
+    {
+        if (!$transformer instanceof Bidirectional) {
```
andrewdalpino: Should we typehint this instead?

```php
public function reverse(Bidirectional $transformer) : self
```

andrewdalpino (commented):
Looks pretty on point @torchello

I left a couple comments regarding the changes to Dataset

Also, I'm not sure if we should introduce this in 1.0 or in 1.1. Leaning toward putting it in the next feature release.

P.S. I wonder, are all transformations reversible?

src/Datasets/Dataset.php
torchello: Makes sense. Also covers some of the scenarios when `reverseTransform` is called before `transform`.

torchello:
> Also, I'm not sure if we should introduce this in 1.0 or in 1.1. Leaning toward putting it in the next feature release.

It's BC compatible, so probably ok to add in 1.0. But up to you.

> P.S. I wonder, are all transformations reversible?

Hm... Probably no, but that's why we have `Bidirectional` interface, right? :)

andrewdalpino:
> It's BC compatible, so probably ok to add in 1.0. But up to you.

I intended to follow semantic versioning such that major version updates to include only/mostly breaking changes. If new APIs are added in a major version it's justified if it replaces an API that was removed. So it sounds like this should go in the next feature update since it does not require breaking changes and is completely new functionality.

andrewdalpino:
> > P.S. I wonder, are all transformations reversible?

> Hm... Probably no, but that's why we have Bidirectional interface, right? :)

It seems as a general rule, if there is no information loss through the transformation, then the transformation is reversible.

Here are two examples in which the transformation is not reversible.

1. [Word Count Vectorizer](https://github.com/RubixML/ML/blob/master/src/Transformers/WordCountVectorizer.php) - token adjacency information is dropped in the transform so there would be no way to reconstruct the sequence of words in a blob of text from only their token counts.

2. [Token Hashing Vectorizer](https://github.com/RubixML/ML/blob/master/src/Transformers/TokenHashingVectorizer.php) - Hash functions are one-way, thus no way to reconstruct the token from its hash.

I'm sure other examples exist. In either case, I think the design of the API is flexible enough to accommodate these differences.

src/Datasets/Dataset.php
torchello: Looks good to me. I will update. What about interface name?

PR #186, opened 2021-07-16
Introduce `ScoreInput` object to aggregate the input data for `Metric…

torchello:
…::score()`` method


andrewdalpino (commented):
This change creates a backwards compatibility break and would need to go in the next major update (if we implement it).

torchello:
Right, I'll close this one for now... Let's catch up re alternative solutions.

PR #201, opened 2022-01-05
Add INF/NAN converter

torchello:
### Problem
[RubixML Client](https://github.com/RubixML/Client) is currently not able to send requests with samples that contain `INF` and `NAN` constants as `json_encode` function does not support serialization of these values.

### Workaround
I suggest adding a new reversible transformer that would be used to convert the constants to some arbitrary values on the client side. Then the server will decode the payload and reverse apply the transformation to get constants back.

### Pull requests
- [x] Add new transformer to `ML` (this PR)
- [ ] Apply transformation on the [Client](https://github.com/RubixML/Client) side
- [ ] Apply reverse transformation on the [Server](https://github.com/RubixML/Server) side

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+- 1.4.0
+  - Added reversible converter for INF/NAN constants
+    
 - 1.3.2
     - Optimize Binary output layer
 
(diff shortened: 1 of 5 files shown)


torchello:
Closing in favor of #202.

PR #202, opened 2022-01-09
Reversible numeric string converter

andrewdalpino:


src/Transformers/NumericStringConverter.php
```diff
@@ -14,21 +14,12 @@
...
  */
-class NumericStringConverter implements Transformer
+class NumericStringConverter implements Transformer, Reversible
 {
-    /**
-     * The numeric string representation of NaN.
-     *
-     * @var string
-     */
-    public const NAN_PLACEHOLDER = 'NAN';
```
torchello, line 30: Should the removal of this constant be mentioned in the CHANGELOG as a BC break?

torchello (approved):
Looks good, thanks! Left one questions about the constant.

src/Transformers/NumericStringConverter.php
andrewdalpino: Yes, good call ... it's a public member of the class so definitely

Issue #207, opened 2022-01-28
Complement Naive Bayes?


DrDub:
Would you be interested in a variant to NB that is more recent and performant?

Rennie, J. D., Shih, L., Teevan, J., & Karger, D. R. (2003). Tackling the poor assumptions of naive bayes text classifiers. In ICML (Vol. 3, pp. 616-623). https://people.csail.mit.edu/jrennie/papers/icml03-nb.pdf

I can give it a try.

andrewdalpino:
Hey @DrDub that would be awesome actually. Let me know if you need any help getting started!

PR #210, opened 2022-02-11
Better visualization of decision trees (fixes #136)

### LINKED ISSUE #136: Better visualization of decision trees (CART)
Currently, our CART implementation offers the `rules()` method which outputs a rules diagram to the terminal that helps the user visualize the tree structure. The diagram uses ASCII characters to represent the nodes of the tree. However, we can go one better by using *actual* graphics. I'm picturing a graph representation with nodes and edges with the root node of the tree at the top and the leaf nodes at the bottom. Each node would be labeled with the column and value it splits on and can include metadata such as the node's purity increase.

I came accross [this](https://github.com/graphp/graphviz) library which may be of service. It's based on Graphviz.

DrDub:
Following the discussion on #136, this change builds on the rules() method from a51d3f197602d624ffdcbc227c2569a0cd1d64f5 to produce a tree description for the graphviz package.

The string generated can be send through the dot command to produce an image.

A test case for is added to ClassificationTreeTest.php

Running this on the dataset from divorce produces an output like
```dot
digraph Tree {
node [shape=box, fontname=helvetica];
edge [fontname=helvetica];
N0 [label="Column_20 < 1"];
N1 [label="Column_52 < 0"];
N2 [label="Column_7 < 0"];
N3 [label="Outcome=divorced",style="rounded,filled",fillcolor=gray];
N2 -> N3;
N4 [label="Outcome=divorced",style="rounded,filled",fillcolor=gray];
N2 -> N4;
N1 -> N2;
...
```
That after running it through `dot -Tpng divorce.dot > divorce.png` renders as:

divorce

Next steps: have a rulesGraphical that produces the image using an extra dependency. (This might be overkill to add a dependency on graphviz for every deployment of RubixML.)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -92,7 +92,10 @@
     "config": {
         "preferred-install": "dist",
         "sort-packages": true,
-        "process-timeout": 3000
+        "process-timeout": 3000,
+        "allow-plugins": {
+            "phpstan/extension-installer": false
+        }
     },
     "funding": [
         {
--- a/composer.json
+++ b/composer.json
@@ -92,10 +92,7 @@
     "config": {
         "preferred-install": "dist",
         "sort-packages": true,
-        "process-timeout": 3000,
-        "allow-plugins": {
-            "phpstan/extension-installer": false
-        }
+        "process-timeout": 3000
     },
     "funding": [
         {
--- a/src/Graph/Trees/DecisionTree.php
+++ b/src/Graph/Trees/DecisionTree.php
@@ -391,10 +391,10 @@ abstract class DecisionTree implements BinaryTree, IteratorAggregate
         $thisNode = $nodesCounter++;
 
         if ($depth == $maxDepth) {
-            $carry .= "N$thisNode [label=\"...\"];" . PHP_EOL;
+            $carry .= "  N$thisNode [label=\"...\"];" . PHP_EOL;
         } elseif ($node instanceof Split) {
             $operator = is_string($node->value()) ? '==' : '<';
-            $carry .= "N$thisNode [label=\"Column_{$node->column()} $operator {$node->value()}\"];" . PHP_EOL;
+            $carry .= "  N$thisNode [label=\"Column_{$node->column()} $operator {$node->value()}\"];" . PHP_EOL;
 
             if ($node->left() !== null) {
                 $this->_rules($carry, $nodesCounter, $maxDepth, $node->left(), $thisNode, 1, $depth);
@@ -404,7 +404,7 @@ abstract class DecisionTree implements BinaryTree, IteratorAggregate
                 $this->_rules($carry, $nodesCounter, $maxDepth, $node->right(), $thisNode, 2, $depth);
             }
         } elseif ($node instanceof Outcome) {
-            $carry .= "N$thisNode [label=\"Outcome={$node->outcome()}";
+            $carry .= "  N$thisNode [label=\"Outcome={$node->outcome()}";
 
             if ($node->impurity() > 0.0) {
                 $carry .= "\\nImpurity={$node->impurity()}";
@@ -413,7 +413,7 @@ abstract class DecisionTree implements BinaryTree, IteratorAggregate
         }
 
         if (!is_null($parentId)) {
-            $carry .= "N$parentId -> N$thisNode";
+            $carry .= "  N$parentId -> N$thisNode";
 
             if ($parentId == 0) {
                 $carry .= ' [labeldistance=2.5, ';
@@ -425,24 +425,6 @@ abstract class DecisionTree implements BinaryTree, IteratorAggregate
                 }
             }
             $carry .= ';' . PHP_EOL;
-        }
-    }
-
-    /**
-     * Return a generator for all the nodes in the tree starting at the root.
-     *
-     * @return \Generator<\Rubix\ML\Graph\Nodes\Decision>
-     */
-    protected function dump() : Generator
-    {
-        $stack = [$this->root];
-
-        while ($current = array_pop($stack)) {
-            yield $current;
-
-            foreach ($current->children() as $child) {
-                $stack[] = $child;
-            }
         }
     }
 }
--- a/src/Graph/Trees/DecisionTree.php
+++ b/src/Graph/Trees/DecisionTree.php
@@ -227,7 +227,7 @@ abstract class DecisionTree implements BinaryTree, IteratorAggregate
                         $current = $current->right();
                     }
                 } else {
-                    if ($sample[$current->column()] < $value) {
+                    if ($sample[$current->column()] <= $value) {
                         $current = $current->left();
                     } else {
                         $current = $current->right();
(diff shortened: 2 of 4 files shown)


DrDub:

andrewdalpino:
Nice @DrDub! This looks great. I'm wondering if a better name for this method now is something like `exportGraphviz()` or `exportDotString()` or whatever. Another idea I had was to have the method accept a "header" array consisting of a map of feature titles to their column offset in the dataset. With that, instead of saying `column_4 < 0.65`, we could say `Height < 0.65`. Thoughts?

Also, I retargeted this at the 2.0 release since it's planned for next month and we don't intend to do another feature release before then.

composer.json
```diff
@@ -92,7 +92,10 @@
     "config": {
         "preferred-install": "dist",
         "sort-packages": true,
-        "process-timeout": 3000
+        "process-timeout": 3000,
+        "allow-plugins": {
+            "phpstan/extension-installer": false
```
andrewdalpino: Just curious why this was needed. BTW we'll fix the stuff Stan is complaining about in another commit. Usually after a while he starts to find more stuff for us to fix.

src/Graph/Trees/DecisionTree.php
```diff
@@ -287,6 +287,34 @@ public function getIterator() : Generator
...
+     * Print a representation of the decision tree suitable to render with the
+     * graphviz tool. For example, writing it to graph.dot then executing:
+     *
+     * dot -Tpng graph.dot
+     *
+     * @param int $maxDepth
+     * @throws RuntimeException
+     * @return string
+     */
+    public function rules(int $maxDepth = -1) : string
```
andrewdalpino: I like the addition of `maxDepth` here to constrain the output

src/Graph/Trees/DecisionTree.php
```diff
@@ -335,4 +363,68 @@ protected function splitImpurity(array $groups) : float
...
+        BinaryNode|Split|Decision $node,
+        ?int $parentId = null,
+        ?int $leftRight = null,
+        int $depth = 0
+    ) : void {
+        ++$depth;
+
+        $thisNode = $nodesCounter++;
+
+        if ($depth == $maxDepth) {
```
andrewdalpino: Very minor but strict comparison here is possible since both `$depth` and `$maxDepth` are integer variables. Strict comparison is sightly more performant. See https://stackoverflow.com/questions/2401478/why-is-faster-than-in-php.

andrewdalpino:
As far as bringing Graphviz in as a dependency, I think it's perfectly fine to throw a runtime exception if Graphviz is not detected on the system when calling the method. We could use a specification to handle this if you want. In our architecture, Specification objects are used for runtime validation. Check out https://github.com/RubixML/ML/blob/master/src/Specifications/ExtensionIsLoaded.php for example. Do you know if  the "dot" syntax changes a lot between Graphviz versions? Thoughts?

Also, I like having one method to output the "dot" format and then a higher-order method that runs Graphviz on the output and pops up an image. That way the user could run Graphviz with their own options if they wanted. Thoughts?

DrDub:
> I'm wondering if a better name for this method now is something like `exportGraphviz()` or `exportDotString()` or whatever. 

Yes, I was thinking in similar lines, just didn't want to stride too far from the original issue. `exportGraphviz()` sounds fine.

> Another idea I had was to have the method accept a "header" array consisting of a map of feature titles to their column offset in the dataset. With that, instead of saying `column_4 < 0.65`, we could say `Height < 0.65`. Thoughts?

That sounds great and I can add it to this PR. This type of exports can get very big full of options and tons of imperative code building strings, etc. Would it make sense to move it to a separate file? I am afraid it is unrelated code to the core of the tree functionality and it will be a distraction to people trying to understand the class.

> Also, I retargeted this at the 2.0 release since it's planned for next month and we don't intend to do another feature release before then.

Makes sense.

composer.json
DrDub: my bad, I didn't intend to check in this change. I'll remove it.

src/Graph/Trees/DecisionTree.php
DrDub: Makes sense. Will do.

DrDub:
> As far as bringing Graphviz in as a dependency, I think it's perfectly fine to throw a runtime exception if Graphviz is not detected on the system when calling the method. 

Sounds good.

> We could use a specification to handle this if you want. In our architecture, Specification objects are used for runtime validation. Check out https://github.com/RubixML/ML/blob/master/src/Specifications/ExtensionIsLoaded.php for example. 

I'll check it in detail when I look into doing the Graphviz bit and closing #136.

> Do you know if the "dot" syntax changes a lot between Graphviz versions? Thoughts?

No, it doesn't. It has been around for 20 years and it is always the same.

> Also, I like having one method to output the "dot" format and then a higher-order method that runs Graphviz on the output and pops up an image. That way the user could run Graphviz with their own options if they wanted. Thoughts?

Yup, this is what I was going for. Particularly because the "dot" format is much more useful than the image. It can generate SVG for example and it can be further embedded into other graphs, etc.

DrDub:
I'll do now some of the changes and we can continue the discussion. When we're happy with it, I can squash the commits. 

We can also not merge this and wait until I finish the whole #136. Just let me know how you want to proceed.

DrDub:
This is how it looks now with the feature names:

divorce

andrewdalpino:
NIce, I think the graphic looks great, much easier for a human to interpret. The one thing we would want I think is to truncate the feature title if it gets "too" long. Maybe the same could be said for the "outcome." Also, I think we could drop the `Outcome=` part as that may be implied already and just increases the amount of visual noise in the graphic. Thoughts?

Lastly, the `I know my spouses' basic anxieties < 0` seems a bit off. From what I understood about this dataset is that it uses the Likert scale (0, 5). Should it be `<= 0`?

I'll keep an eye on the PR but so far everything looks great! Let me know when it's ready to merge. I think it would be best to get everything into one PR but if you prefer to do it in multiple PRs we could use a feature branch to hold the changes.

DrDub:
Okay, I'm getting there. Still thinking if you want so much fine tuning this functionality might be better off in a separate file.

> This type of exports can get very big full of options and tons of imperative code building strings, etc. Would it make sense to move it to a separate file? I am afraid it is unrelated code to the core of the tree functionality and it will be a distraction to people trying to understand the class.

DrDub:
> Lastly, the I know my spouses' basic anxieties < 0 seems a bit off. From what I understood about this dataset is that it uses the Likert scale (0, 5). Should it be <= 0?

It seems we have found a bug in the decision tree code :-)

The code that computes the split (Labeled.php#595) uses <= buuuut the code that uses the split

DecisionTree.php#230 (function search) uses <

I'll fix that line and change the printout to <=

DrDub:
Now that I added a function to export to an image (on disk). I looked at graphp/graphviz but all I'd be using from there would be a function (the one that calls 'dot') so I just recoded that here and saved a dependency.

I'd rather have a function that returns the bytes of the image but that seems to be difficult to accomplish in PHP in an operating system independent manner.

This how it looks now:

divorce

This is ready to merge on my end but will be happy to make any changes, just let me know. If we can merge this I can start looking into #207.

andrewdalpino:
> Okay, I'm getting there. Still thinking if you want so much fine tuning this functionality might be better off in a separate file.

Sure, I'm down. Where would you put it? Would it be reusable by other components or is it only applicable to the Decision Tree?

> It seems we have found a bug in the decision tree code :-)

Excellent catch! I'll work on a bugfix for master and any other maintenance branches this may exist in.

> Now that I added a function to export to an image (on disk). I looked at graphp/graphviz but all I'd be using from there would be a function (the one that calls 'dot') so I just recoded that here and saved a dependency.

I think that was a good choice.

> This is ready to merge on my end but will be happy to make any changes, just let me know.

For sheezy! I'll give this another quick look ASAP. Thanks!

DrDub:
I have been thinking maybe I should open a bug for the decision tree and have a PR with a test case for it, so people can backtest on their implementation for a fix. What do you think?

PR #221, opened 2022-03-26
Multiclass Brier Score

andrewdalpino:
See https://en.wikipedia.org/wiki/Brier_score


tests/CrossValidation/Metrics/BrierScoreTest.php
```diff
@@ -0,0 +1,141 @@
...
+                [1 => 0.0, 2 => 1.0, 3 => 0.0],
+                [1 => 1.0, 2 => 0.0, 3 => 0.0],
+            ],
+            [3, 2, 1],
+            0.0,
+        ];
+
+        yield [
+            [
+                [1 => 0.2, 2 => 0.2, 3 => 0.0],
```
torchello, line 115: Are these probabilities correct? (0.4 in total)

> Metric assumes probabilities are between 0 and 1 and their joint distribution sums to 1.

Should we have a specification to validate a sum?

torchello (commented):
Please see my question.

tests/CrossValidation/Metrics/BrierScoreTest.php
andrewdalpino: That's a good idea but I think it would add too much overhead (we'd have to check every joint probability distribution array). This was a typo on my part.

PR #225, opened 2022-04-03
Top K Accuracy

andrewdalpino:


torchello (approved):
Nice. What would be a real-life use case for this metric?

andrewdalpino:
Like say, it was ok for a prediction to be in the top 3 categories or top 5

PR #228, opened 2022-05-17
GraphViz Helper

andrewdalpino:


src/Graph/Trees/DecisionTree.php
```diff
@@ -468,26 +462,60 @@ protected function _exportGraphviz(
...
+                $carry .= "\\nImpurity={$impurity}";
             }
-            $carry .= '",style="rounded,filled",fillcolor=gray];' . PHP_EOL;
+
+            $carry .= '"';
+
+            if (is_string($outcome)) {
+                $hash = crc32($outcome) % self::MAX_COLORS;
+
+                $fillColor = substr('00000' . dechex($hash), -6);
```
andrewdalpino: Now that I think about it we could also generate the hash as a hex string and then just truncate it beyond the first 24 bits

DrDub (approved):
LGTM

For the next iteration, we can move all the color / truncation / tree generation to an exporting helper.

andrewdalpino:
> For the next iteration, we can move all the color / truncation / tree generation to an exporting helper.

What do you think about converting the recursion into iteration? I think that might clean up the logic a bit and may be a tid faster. I remember trying to come up with the iterative version before but I ended up empty handed. Maybe it's time for another go.

PR #235, opened 2022-06-03
Graph subsystem optimizations and refactor

andrewdalpino:
This refactor makes it more efficient to serialize and store graph nodes as well as the splitting of the BinaryNode interface into BinaryNode and HasBinaryChildren interfaces which simplifies leaf node implementations.

The storage efficiency comes primarily from eliminating unused class properties.


andrewdalpino:
Unfortunately, since some class properties of various leaf nodes have been removed, the classes will now hash to a different revision number than ones that were saved with a previous version of the library. For this reason, I'm targetting this at the 2.1 release, rather than the current bugfix release as to not "shock" people that aren't expecting their models not to work after upgrading.

tests/Graph/Nodes/SplitTest.php
```diff
@@ -114,4 +113,21 @@ public function n() : void
     {
         $this->assertSame(self::N, $this->node->n());
     }
+
+    /**
+     * @test
+     */
+    public function cleanup() : void
```
DrDub, line 120: Here you seem to be testing the cleanup behaviour of the nodes. But initially it reads as a cleanup for the test case. So "testCleanup" might be less confusing. Second, the functionality you're testing makes sense to test but why is it in the tests for Split rather than tests for nodes?

src/Graph/Nodes/BinaryNode.php
```diff
@@ -16,70 +14,9 @@
...
-     * @return int
-     */
-    public function balance() : int;
-
-    /**
-     * Set the left child node.
-     *
-     * @param self $node
-     */
-    public function attachLeft(BinaryNode $node) : void;
```
DrDub, line 60: so you're dropping the attach/detach functionality. The attach is dropped from the interface but not from the trait. The detach is completely gone. I guess the detach was not being used so it sounds good. Why dropping the attach from the interface while keeping it in the trait? Just trying to understand where you want to take the design.

src/Graph/Nodes/BinaryNode.php
```diff
@@ -16,70 +14,9 @@
...
-     * @return \Traversable<\Rubix\ML\Graph\Nodes\BinaryNode>
-     */
-    public function children() : Traversable;
-
-    /**
-     * Return the left child node.
-     *
-     * @return \Rubix\ML\Graph\Nodes\BinaryNode|null
-     */
-    public function left() : ?BinaryNode;
```
DrDub, line 30: What's wrong about keeping left() / right() nodes for a BinaryNode...?

DrDub (approved):
The bits from Graphviz are giving now some errors that I guess will get addressed when #228 gets merged. My other questions are just comments. Seems good to merge.

tests/Graph/Nodes/SplitTest.php
andrewdalpino: Yeah we use PHPUnit annotations rather than prefixing the test method with the word `test` so for consistency I'd rather stick to the plan, although, I agree one who may be unfamiliar with PHPUnit might mistake this for some sort of method that resets or cleans up stuff left over from a test.

As for why the unit test lives in the Split class test, no reason other than we needed somewhere to put it. Perhaps I'll add this test to the rest of the nodes that implement this method as well.

src/Graph/Nodes/BinaryNode.php
andrewdalpino: Good observation. The `attach()` API was moved to the `HasBinaryChildren` interface since it is possible for some `BinaryNode`s (ex. leaf nodes) to not have children.

src/Graph/Nodes/BinaryNode.php
andrewdalpino: Certain `BinaryNode`s such as leaf nodes do not have any children, by definition. The BinaryNode interface now reflects this. We have a separate interface `HasBinaryChildren` for BinaryNodes that *do* have children.

andrewdalpino:
Thanks for the review @DrDub! I added some more tests and fixed a missing exception class import. I believe the issues with Stan are resolved in the Graphviz helper PR.

PR #239, opened 2022-07-10
Added new image rotator transformer.

Boorinio:
Hello again,
This pr introduces the Random image rotation transformer, a very important one imo. The constructor gets the maximum degrees of rotation and then it applies it to the gd image while keeping the same center and cropping the rest of the image to fit the initial width and height. Here is an example on how it works on the ISIC dataset.

From left to right:
Original Image, rotated image, cropped image to fit the original size
image


src/Transformers/ImageRandomRotationer.php
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * Rotates the image between 0 and a given number of degrees
+ *
+ * > **Note**: The GD extension is required to use this transformer.
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Stylianos Tzourelis
+ */
+class ImageRandomRotationer implements Transformer
```
andrewdalpino: I think just `ImageRotator` would be sufficient as well

src/Transformers/ImageRandomRotationer.php
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Return the string representation of the object.
+     *
+     * @internal
+     *
+     * @return string
+     */
+    public function __toString() : string
+    {
+        return "Image Random Rotationer with a maximum of {$this->degrees} degrees.";
```
andrewdalpino: See other classes of this type for how the string form is formatted. This is inconsistent.

andrewdalpino (commented):
Hey this is so cool! I gave it a quick glance and will give it a full review ASAP. Thanks :)

src/Transformers/ImageRandomRotationer.php
Boorinio: You are right! Thanks, I wasn't a fan of the previous name either :P

src/Transformers/ImageRandomRotationer.php
Boorinio: Done!

DrDub:
I have reviewed the PR and run the tests locally after merging it with current master. It all looks good.

Image rotation is a particular case of applying a rigid transformation to the image. We can start with a rotation and leave other transformations for other classes or extend this class as needed.

My only comment is the use of the color black as the background to fill the empty parts. First, using "0" in the code to mean black is a little hard to grasp (I had to read up on imagerotate to see it). Second, this color might be worth parameterizing. Other options include using the color in the "border" of the image, represented, for example, by the pixel in the upper left corner. But this might all be just feature creep.

Let us see what Andrew has to say. Very nice PR!

andrewdalpino:
Hey @Boorinio and @DrDub sorry it took me so long ... Yeah I agree super cool feature and I think the implementation is on point. I could go either way on parameterizing the "fill" color, for now we could leave it out and then parameterize it later without breaking backwards compatibility (default value will be vanta black (000)).

andrewdalpino (approved):

andrewdalpino:
We'll release this in 2.2!

PR #242, opened 2022-09-14
Gower distance

andrewdalpino:
This is the Gower distance kernel - a dual type, NaN-safe distance kernel that is pretty slow relative to other distance kernels but makes up for it by accepting pretty much any type of input including samples with missing values.

### DIFF
--- a/src/Kernels/Distance/Gower.php
+++ b/src/Kernels/Distance/Gower.php
@@ -1,0 +1,122 @@
+<?php
+
+namespace Rubix\ML\Kernels\Distance;
+
+use Rubix\ML\DataType;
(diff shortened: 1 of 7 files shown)


torchello (approved):
🔥

src/Kernels/Distance/Gower.php
```diff
@@ -0,0 +1,122 @@
...
+
+                    break;
+
+                case is_float($valueB) and is_nan($valueB):
+                    ++$nn;
+
+                    break;
+
+                case !is_string($valueA) and !is_string($valueB):
+                    $distance += abs($valueA - $valueB)
```
andrewdalpino, line 92: This is where the `range` parameters comes into play

PR #244, opened 2022-09-15
One Vs Rest Classifier

andrewdalpino:
One Vs Rest is an ensemble learning technique that trains one binary classifier for each potential class label.


src/Classifiers/OneVsRest.php
```diff
@@ -0,0 +1,254 @@
...
+            $dist = [];
+
+            foreach ($votes as $j => $proba) {
+                $dist[$classes[$j]] = $proba['y'];
+            }
+
+            $total = array_sum($dist);
+
+            foreach ($dist as &$probability) {
+                $probability /= $total;
```
DrDub, line 234: What you are doing here looks like it should be a SoftMax. Did this formula come from some paper. My gut feeling is averaging the probabilities shouldn't really work. If you want to have this here it might work if you change line 228 to

```
$dist[$classes[$j]] = $proba['y'] > $proba['n'] ? 1.0 : 0.0;
```

But using a SoftMax will work better if you want to use `$proba[y]` in line 228.

DrDub (commented):
Sorry for the late review. The current code will classify correctly. It is the `proba` values that concern me.

PR #245, opened 2022-09-17
Add offset and jitter to Image Rotator

andrewdalpino:
The current implementation adds unidirectional jitter to the image and the offset is always 0 degrees. This change allows the user to set the starting rotation (offset) and adds bidirectional jitter. The idea is that a user may want to do things like ...

```php
use Rubix\ML\Transformers\ImageRotator;

$transformer = new ImageRotator(-90.0, 0.0); // Static rotation of 90 degrees clockwise.

$transformer = new ImageRotator(90.0, 0.0); // Static rotation of 90 degrees anti-clockwise.

$transformer = new ImageRotator(0.0, 0.1); // Apply a slight jitter about the origin.

$transformer = new ImageRotator(180.0, 0.5); // Flip the image upside down, and apply moderate jitter.

$transformer = new ImageRotator(0.0, 1.0); // Randomly rotate about the full 360 degrees. The offset doesn't make a difference in this scenario.
```

I think this aligns more with how users will use this Transformer in practice. That is, to permute images to make their models more robust.


andrewdalpino:
One thing I forgot to implement is "wrapping" ... I'll do that later

PR #248, opened 2022-09-20
Column Filter Extractor

andrewdalpino:
Similarly to Column Pick (https://docs.rubixml.com/2.0/extractors/column-picker.html) this Extractor iterates over a subset of the base iterator. However, unlike Column Picker, which selects the specified columns, this one filters them and selects the rest of the columns.

I found this to be useful when filtering columns like `ID` and `created_at` fields.


DrDub (commented):
All good.

PR #250, opened 2022-09-25
2.2

andrewdalpino:
Version 2.2 here


src/Classifiers/OneVsRest.php
```diff
@@ -0,0 +1,254 @@
...
+        $probabilities = [];
+
+        foreach ($aggregate as $votes) {
+            $dist = [];
+
+            foreach ($votes as $j => $proba) {
+                $dist[$classes[$j]] = $proba['y'];
+            }
+
+            $total = array_sum($dist);
```
andrewdalpino, line 233: We could also potentially distribute the 'n' probabilities over the remaining classes so that the distribution still adds to 1 and there would be no need for this `array_sum` (O(n)) as we'd just divide over the number of binary classifiers instead.

src/Transformers/ImageRotator.php
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * Randomly rotates the image between 0 and a given number of max degrees.
+ *
+ * > **Note**: The GD extension is required to use this transformer.
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Stylianos Tzourelis
+ */
+class ImageRotator implements Transformer
```
andrewdalpino, line 24: I haven't had time to test this class yet using actual images. It would be awesome if someone had the chance to test this.

src/Classifiers/OneVsRest.php
```diff
@@ -0,0 +1,254 @@
...
+            $subset->transformLabels($binarize);
+
+            $task = new TrainLearner($estimator, $subset);
+
+            $this->backend->enqueue($task);
+        }
+
+        $classifiers = $this->backend->process();
+
+        $classifiers = array_combine($classes, $classifiers) ?: [];
```
Jeckerson, line 175:
```suggestion
        $classifiers = array_combine($classes, $classifiers);
```

benchmarks/Kernels/Distance/GowerBench.php
```diff
@@ -0,0 +1,68 @@
...
+    protected $bSamples;
+
+    /**
+     * @var \Rubix\ML\Kernels\Distance\Gower
+     */
+    protected $kernel;
+
+    public function setUp() : void
+    {
+        $generator = new Blob([0, 0, 0, 0, 0, 0, 0, 0], 5.0);
```
andrewdalpino, line 34: I think these data should be normalized as well, since it's hard to know the range of these values. All we know is that they are centered at zero and have a standard deviation of 5.0.

docs/classifiers/one-vs-rest.md
```diff
@@ -0,0 +1,25 @@
+<span style="float:right;"><a href="https://github.com/RubixML/ML/blob/master/src/Classifiers/OneVsRest.php">[source]</a></span>
+
+# One Vs Rest
+One Vs Rest is an ensemble learning technique that trains one binary classifier for each potential class label.
```
andrewdalpino: I think we could probably say another sentence or two about One Vs Rest to help give the user a richer idea of what it can do

src/Classifiers/OneVsRest.php
andrewdalpino: Unfortunately, this is needed for PHP 7.4 :/

src/Classifiers/OneVsRest.php
andrewdalpino: This actually changes the probabilities (it has a smoothing effect) and makes it less accurate in the unit test so not implementing.

src/Transformers/ImageRotator.php
andrewdalpino: Tested. Works well.

PR #259, opened 2022-10-02
Image Vectorizer Color Channels

andrewdalpino:
This change lets the user choose which color channels to encode.

### DIFF
--- a/benchmarks/Transformers/ImageVectorizerBench.php
+++ b/benchmarks/Transformers/ImageVectorizerBench.php
@@ -33,7 +33,7 @@ class ImageVectorizerBench
 
         $this->dataset = Unlabeled::build($samples);
 
-        $this->transformer = new ImageVectorizer();
+        $this->transformer = new ImageVectorizer(true, true, true);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


benchmarks/Transformers/ImageVectorizerBench.php
```diff
@@ -33,7 +33,7 @@ public function setUp() : void
 
         $this->dataset = Unlabeled::build($samples);
 
-        $this->transformer = new ImageVectorizer();
+        $this->transformer = new ImageVectorizer(true, true, true);
```
DrDub, line 36: This looks very bizarre from a user's perspective (`true, true, true`). The default should be the most common option and without lots of typing.

I can see this change is an improvement over the previous code but "grayscale processing" usually means processing:

```
gray=0.299red+0.587green+0.114blue
```

What about `ImageVectorizer()` -> all 3 channels
`ImageVectorizer(gray: true)` -> the formula above
`ImageVectorizer(green: false)` -> only red and blue

The above is straightforward using named parameters now that we require PHP 8

andrewdalpino: I agree three consecutive booleans looks kinda weird but I'm not sure if you;'re suggestion is more intuitive either. Perhaps we should just leave it the way it is. I can't think of many use cases where you'd want to isolate specific color channels.

PR #261, opened 2022-10-06
Bump flysystem to v2.1.1 or later

ChristophWurst:
This patch allows updating flysystem to newer versions for projects that are still on 0.x.


ChristophWurst:

ChristophWurst:
CI failure is caused by the lack of allowing the phpstan composer plugin. Submit https://github.com/RubixML/ML/pull/262 to fix this for the `0.0` branch.

Issue #260, opened 2022-10-02
Warning: Ambiguous class resolution, `Rubix\ML\Kernels\Distance\Gower`


andrey-helldar:
I'm using the `BM25Transformer` transformer contained in the [`rubixml/extras`](https://github.com/RubixML/Extras) repository, but after updating RubixML to version 2.2, a warning occurs when installing dependencies:

> Warning: Ambiguous class resolution, "Rubix\ML\Kernels\Distance\Gower" was found in both "D:/domains/dragon-code/apps/webhooks/vendor/rubix/extras/src\Kernels\Distance\Gower.php" and "D:/domains/dragon-code/apps/webhooks/vendor/rubix/ml/src\Kernels\Distance\Gower.php", the first will be used.

- https://github.com/RubixML/Extras/blob/master/src/Kernels/Distance/Gower.php
- https://github.com/RubixML/ML/blob/master/src/Kernels/Distance/Gower.php

Same namespace: `Rubix\ML\Kernels\Distance\Gower`

PR: https://github.com/RubixML/ML/pull/242

andrewdalpino:
I think we still need to remove Gower from the Extras package as we moved it over from Extras to the main package for version 2.2. Will do this as soon as I have time.

andrey-helldar:
Thank you

andrewdalpino:
Ok Gower was removed in Extras 2.0.1-beta so now it should be compatible with ML 2.2.0.

https://github.com/RubixML/Extras/releases/tag/2.0.1-beta

andrey-helldar:
@andrewdalpino, thank you!

PR #262, opened 2022-10-06
Allow the phpstan composer plugin

ChristophWurst:
As seen in https://github.com/RubixML/ML/pull/261 CI fails for PRs against `0.0` due to the composer plugin not being allowed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,10 @@
     "config": {
         "preferred-install": "dist",
         "sort-packages": true,
-        "process-timeout": 3000
+        "process-timeout": 3000,
+        "allow-plugins": {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
PHPStan doesn't like this branch it seems.

Before I end up like [this](https://www.youtube.com/watch?v=AbSehcT19u0), let me know if you accept the patch #260 and if green CI is a hard requirement.

andrewdalpino:
Nah, we don't need to be green. This is an older branch and PHPStan has a way of creeping up on you if it hasn't been updated in a while.

andrewdalpino (approved):

PR #263, opened 2022-10-14
Fix Extra Tree Divide By Zero

andrewdalpino:


src/Graph/Trees/ExtraTree.php
```diff
@@ -91,14 +92,18 @@ protected function split(Labeled $dataset) : Split
                 $min = min($values);
                 $max = max($values);
 
-                $phi = $randMax / max(abs($max), abs($min));
+                $maxAbs = max(abs($max), abs($min));
+
+                $phi = $maxAbs != 0.0 ? $randMax / $maxAbs : $randMax;
```
DrDub, line 97: Why `$randMax` when `$maxAbs == 0.0` will work here?

This code tries to select a floating point number between `$min` and `$max` using all the resolution available in the random number generator, right?

DrDub (commented):
Sorry for the late review.

PR #264, opened 2022-10-15
Drop Feature

andrewdalpino:
This feature adds a method to the Dataset object API for dropping a single feature column from the dataset. I came into a need for this when writing the Churn tutorial. Specifically, we wanted to extract the customer ID from the database for use later but we don't need the ID for training or inference.


DrDub (commented):
Very good!

PR #273, opened 2023-01-22
Emoji remover transformer

Boorinio:
Hello again!

This is another transformer I used in a personal project of mine, I hope you find it useful enough to merge!
I also hope I can find some time in the near future to contribute some primitives!

Best Regards and thanks for your hard work!


andrewdalpino (commented):
Hey this is pretty handy, thanks @Boorinio. I think instead of implementing an entire Transformer for this, we could probably just add a new constant to the RegexFilter Transformer that matches emojis.

See https://github.com/RubixML/ML/blob/master/src/Transformers/RegexFilter.php#L34 for example.

If we go this route, we could use the existing test but add additional assertions for emojis.

Boorinio:
> Hey this is pretty handy, thanks @Boorinio. I think instead of implementing an entire Transformer for this, we could probably just add a new constant to the RegexFilter Transformer that matches emojis.
> 
> See https://github.com/RubixML/ML/blob/master/src/Transformers/RegexFilter.php#L34 for example.
> 
> If we go this route, we could use the existing test but add additional assertions for emojis.

Hey again,
I completely missed this transformer!! I updated the pr, thanks for the response!!

andrewdalpino (approved):
Dope

PR #286, opened 2023-04-15
Update Sentence.php and SentenceTest.php

BahmanMehrdad:
The problem of using other languages such as Farsi and Arabic has been solved.

### DIFF
--- a/src/Tokenizers/Sentence.php
+++ b/src/Tokenizers/Sentence.php
@@ -20,7 +20,7 @@ class Sentence implements Tokenizer
      *
      * @var string
      */
-    protected const SENTENCE_REGEX = '/(?<=[.?!])\s+(?=[a-z])/i';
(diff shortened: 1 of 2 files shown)


andrewdalpino:
This is fantastic @BahmanMD thank you. I just want to do a bit more testing with other languages to make sure that we didn't break anything by making this change. If this change does not break backward compatibility then we'll release it with the next feature update, otherwise we'll have to release it with ML 3.0.

andrewdalpino:
Ok, I also want to see how this affects performance since the regular expression used to match the sentences is more complex now.

andrewdalpino:
Not sure why GitHub closed this PR, maybe because the target branch was deleted, but I can't figure out how to unclose it. You may have to resubmit a new PR and target the master branch.

BahmanMehrdad:
@andrewdalpino  Sure.

andrewdalpino:
Thank you @BahmanMD

PR #294, opened 2023-05-16
Added ability to set the escape character in CSV extractor.

Boorinio:
Hello again,
A lot of programs use different separators for csv (I personally encountered after I tried to import a csv from tableplus) The default for php is a backslash but it would be nice to be able to set, it this pr solves that.

### DIFF
--- a/docs/extractors/csv.md
+++ b/docs/extractors/csv.md
@@ -9,18 +9,19 @@ A plain-text format that use newlines to delineate rows and a user-specified del
 **Interfaces:** [Extractor](api.md), [Writable](api.md)
 
 ## Parameters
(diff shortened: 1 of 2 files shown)


andrewdalpino (approved):
Awesome, looks good! I think we need to validate that the escape char is non empty to appease Stan though,

Boorinio:
> Awesome, looks good! I think we need to validate that the escape char is non empty to appease Stan though,

Thanks for the fast reply, fixed!

andrewdalpino:
Looks good! We can Ignore PHP 8.1. on MacOS CI run since it's not our issue. Thanks again!

PR #296, opened 2023-05-22
Added map in Dataset and tests for it.

Boorinio:
Fixed 
https://github.com/RubixML/ML/issues/295


andrewdalpino:
Hey @Boorinio it's so funny we were just talking about this in our chat. Map() was removed a while back but we forgot to remove the docs. The Lambda function Transformer replaces this method and is a little more flexible. Having that said, the fix was not to reimplement the map() method but to remove the docs. Thanks again and appreciate the submission!

PR #299, opened 2023-05-29
Blob Simulator

andrewdalpino:
This adds a methos `simulate()` to the Blob generator allowing you to pass in a Dataset object and it will build a Blob generator that generates samples from the same Gaussian distribution as the Dataset object.


Boorinio (approved):
👏

andrewdalpino:
Thanks for the review @Boorinio!

PR #314, opened 2024-01-16
Wrapper interface

ElGigi:
Added a `Wrapper` interface to identify enclosing models.

And thus allow for example to test if a included model implements a particular interface like `LoggerAwareInterface`.


andrewdalpino:
Cool! Thanks @ElGigi, I think we had something like this before but I can't remember why we got rid of it! I'll give it a closer look when I can.

src/Wrapper.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Rubix\ML;
+
+/**
+ * Wrapper
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Andrew DalPino
```
andrewdalpino: Shouldn't this be your name and not my name?

src/Wrapper.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+interface Wrapper extends Estimator
```
andrewdalpino: I'm not 100% sold on the name of this interface ... on one hand, if we only ever have one type of "wrapper" object then this name would be perfect, but on the other hand, if we implement different types of wrappers then ModelWrapper or EstimatorWrapper or something like that would be more appropriate. Let me think about this one for a little bit. I'm leaning toward keeping it the way it is unless I can find a compelling use case for another type of wrapper.

andrewdalpino (commented):
Ok I left some comments ... I remember having an interface like this before, but I can't remember why we removed it exactly. My guess is that it didn't serve any internal use cases so we just dropped it.

I can think of a few external use cases for an interface like this. The one you mentioned is a good one. Do you have any in particular in your own projects?

andrewdalpino:
P.S. To fix the coding style automatically you can run `compose fix` from the project root

src/Wrapper.php
ElGigi: Done 👍

src/Wrapper.php
ElGigi: Done 👍

ElGigi:
> Ok I left some comments ... I remember having an interface like this before, but I can't remember why we removed it exactly. My guess is that it didn't serve any internal use cases so we just dropped it.
> 
> I can think of a few external use cases for an interface like this. The one you mentioned is a good one. Do you have any in particular in your own projects?

In our project, we abstract of the final model. To allow to define own model in sub-libraries.

The final service don't known the model but only interface, so to define if we set a logger or if the model is trainable, we need to retrieve the base estimator.

ElGigi:
> P.S. To fix the coding style automatically you can run `compose fix` from the project root

I do 👍 

I didn't want to do that because it would modify too many files. Unrelated to my PR.

PR #315, opened 2024-02-07
Ridge regressor: switch to numpower

mcharytoniuk:
Thoughts so far:

1. it's easy to build and run NumPower
2. when using GPU got some SIGSEGV when transposing a matrix, other operations seem to be fine
3. even on CPU there is a performance gain
4. to avoid switching between PHP context and internal Matrix representation in NumPower it might be necessary to implement some more matrix operations in NumPower?


andrewdalpino:
Very nice work! We're on the same page. For the changes we need to Num Power, let's temporarily fork it to test the POC and then submit any PRs to the main repo when it's all done.

I'll continue to try to hunt down Henrique lol :)

andrewdalpino (approved):

PR #316, opened 2024-02-07
move tools to a separate directory

mcharytoniuk:
The goal is to avoid dependencies between tools. 

For example the latest `phpstan` and the latest `phpunit` depend on a different version of https://github.com/sebastianbergmann/diff 

Also it won't cause as many potential dependency conflicts between RubixML and the end user projects.

I modified `composer.json`, so scripts should work exactly as before (for example `composer anayze` runs `phpstan`).

I made an exception for `phpunit`, so `phpstan` can still analyze the test files.

Next step would be to also add https://github.com/rectorphp/rector to the tools directory and update phpunit to v11 (tbd in a different PR)


andrewdalpino:
Is this solving a problem that we have currently? Or a potential one in the future?

mcharytoniuk:
@andrewdalpino it solves the current problem: I can't use the latest `PHPUnit` version alongside the latest `PHPStan` because they conflict with the required https://github.com/sebastianbergmann/diff library version. 

I want to upgrade `PHPUnit`, because the current version reaches EOL in the February (at least end of bugfix support): https://phpunit.de/supported-versions.html

I'm just stuck in a dependency hell - I wanted to decouple all those tooling dependencies from the core codebase. :D

andrewdalpino (approved):
As long as this doesn't effect the installability via `composer require rubix/ml` then I think we're good to go

andrewdalpino:
Hey @mcharytoniuk is this still needed?

stalkerboyam (approved):

mcharytoniuk:
@andrewdalpino Yeah, it was supposed to avoid conflcits between the primary `composer.json` and unnecessary dependencies between development tools and the Rubix framework itself.

PR #324, opened 2024-02-14
Convert blob generator to numpower

### LINKED ISSUE #321: Convert `Generators\Blob` to NumPower

mcharytoniuk:


andrewdalpino:
Hey @mcharytoniuk I know we spoke about this on Telegram too, but I just wanted to mention it here too that we'll probably favor usability over performance with these Generator classes so likely will not need this for ML 3.0 but we'll keep it here just in case.

PR #327, opened 2024-03-23
Swoole fix `setAffinity` call on Mac

mcharytoniuk:


src/Backends/Swoole.php
```diff
@@ -98,7 +98,10 @@ function (Process $worker) use ($maxMessageLength, $queueItem) {
                 true,
             );
 
-            $workerProcess->setAffinity([$currentCpu]);
+            if (method_exists($workerProcess, 'setAffinity')) {
+                $workerProcess->setAffinity([$currentCpu]);
```
andrewdalpino, line 102: So this looks like it's telling the worker process to bind to a particular unit. Is that right? Does this provide some performance enhancement? Does it still work as intended without setting the CPU affinity?

andrewdalpino (approved):
Looks good, just wanted to confirm a couple things before we merge. See comment

src/Backends/Swoole.php
mcharytoniuk: It does work as intended; yeah  it will be slower on mac because I think it will cause more context switching.

I added that on top of the other changes, so it wasn't the primary driver behind the performance gains.

PR #331, opened 2024-04-29
Add numpower to the GitHub workflow pipeline

tanmayk:
https://github.com/RubixML/ML/issues/325

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,10 @@ jobs:
           extensions: svm, mbstring, gd, fileinfo, swoole
           ini-values: memory_limit=-1
 
-      - name: Install OpenBLA
+      - name: Install OpenBLAS
         run: |
-          git clone https://github.com/OpenMathLib/OpenBLAS.git
-          cd OpenBLAS
-          make
-          make install
+          apt-get update -q
+          apt-get install -qy libopenblas-dev
 
       - name: Install NumPower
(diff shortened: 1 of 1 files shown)


tanmayk:

andrewdalpino:
Nice, @tanmayk thank you! I'm curious, for the OpenBLAS dependency, could we install that from a repo such as the Debian one like we do with the Tensor extension?

https://github.com/RubixML/Tensor/blob/master/.github/workflows/ci-ext.yml#L62

.github/workflows/ci.yml
```diff
@@ -23,6 +23,23 @@ jobs:
           extensions: svm, mbstring, gd, fileinfo, swoole
           ini-values: memory_limit=-1
 
+      - name: Install OpenBLA
```
andrewdalpino: "Install OpenBLAS" - add an 'S' here

tanmayk: Fixed. Thanks.

tanmayk:
@andrewdalpino Thank you for the review. Yeah you are right. I've updated the PR accordingly.

tanmayk:
@andrewdalpino Any thoughts? If this gets merged, all [these](https://github.com/RubixML/ML/issues?q=is%3Aissue+is%3Aopen+label%3Aoptimization+-label%3A%22help+wanted%22+) issues can be tested with pipeline.

PR #332, opened 2024-05-10
Dual versions for parallel

### LINKED ISSUE #305: Requirements not resolved to an installable set of packages
Hi there,

It seems as if `rubix/ml` can't be installed with either `amphp/amp ^3` or `amphp/phpunit-util ^3` because of `amphp/parallel`. 

```
composer install
Deprecation Notice: Using ${var} in strings is deprecated, use {$var} instead in phar:///usr/local/bin/composer/src/Composer/Autoload/AutoloadGenerator.php:879
Deprecation Notice: Using ${var} in strings is deprecated, use {$var} instead in phar:///usr/local/bin/composer/src/Composer/Autoload/AutoloadGenerator.php:884
Composer is operating significantly slower than normal because you do not have the PHP curl extension enabled.
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Loading composer repositories with package information
Info from https://repo.packagist.org: #StandWithUkraine
Updating dependencies
Deprecation Notice: Using ${var} in strings is deprecated, use {$var} instead in phar:///usr/local/bin/composer/src/Composer/DependencyResolver/Problem.php:366
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - amphp/parallel[v1.3.0, ..., v1.4.3] require amphp/amp ^2 -> found amphp/amp[v2.0.0, ..., v2.6.2] but it conflicts with your root composer.json require (^3).
    - chesslablab/php-chess dev-master requires rubix/ml ^2.4 -> satisfiable by rubix/ml[2.4.0].
    - rubix/ml 2.4.0 requires amphp/parallel ^1.3 -> satisfiable by amphp/parallel[v1.3.0, ..., v1.4.3].
    - Root composer.json requires chesslablab/php-chess dev-master -> satisfiable by chesslablab/php-chess[dev-master].
```

Thanks for the help,

27pchrisl:
Hi @andrewdalpino,

To resolve #305 is it suitable to dual-require parallel versions 1.3 and 2.2, and let Composer choose the right one based on your PHP version?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "require": {
         "php": ">=7.4",
         "ext-json": "*",
-        "amphp/parallel": "^1.3",
+        "amphp/parallel": "^1.3|^2.2",
         "andrewdalpino/okbloomer": "^1.0",
         "psr/log": "^1.1|^2.0|^3.0",
         "rubix/tensor": "^3.0",
--- a/src/Backends/Amp/CallableTask.php
+++ b/src/Backends/Amp/CallableTask.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Rubix\ML\Backends\Amp;
+
+use Amp\Cancellation;
+use Amp\Parallel\Worker\Task;
+use Amp\Sync\Channel;
+
+/**
+ * @template-covariant TResult
+ * @template TReceive
+ * @template TSend
+ * @implements Task<TResult, TReceive, TSend>
+ */
+class CallableTask implements Task
+{
+    /**
+     * @var callable
+     */
+    private $callable;
+
+    public function __construct(callable $callable)
+    {
+        $this->callable = $callable;
+    }
+
+    public function run(Channel $channel, Cancellation $cancellation) : mixed
+    {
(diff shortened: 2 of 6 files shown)


andrewdalpino:
Hey thanks @27pchrisl this could very well be a solution but we may have to make some changes to the code itself. See the failed test here https://github.com/RubixML/ML/actions/runs/9035283567/job/24858504227?pr=332 and let me know what you think. Thanks again!

27pchrisl:
Hiya @andrewdalpino, I created an `Amp2` class that the package user can use if they have the v2 of `parallel`. Essentially a clone of `Amp` but implementing the newer interface.

27pchrisl:
I could implement a common abstract class, but it would only really abstract away the constructor and the __toString method, so I have created them completely separately.

andrewdalpino:
Unfortuantely we don't have the resources to support this at this time. Will look to upgrading to Amp Parallel 2.2 in ML 3.0. Thank you!

PR #333, opened 2024-05-15
Allow normalizers to skip NaN values

27pchrisl:
Hi,

I have a dataset with a lot of missing data, and I wanted to use the `HotDeckImputer` with the `Gower` kernel to fill in the blanks. I had preprocessed my dataset so that `null`s were converted to `NAN` or `?` depending on the data type.

The problem was then that `Gower` expects continuous features to have been normalized. I then wanted to use the `MinMaxNormalizer` to do this on the continuous features in the dataset, but it doesn't handle `NAN` - essentially every value is normalized to zero.

I updated the `MinMaxNormalizer` and the `MaxAbsoluteScaler` to skip `NAN` values, and compute min/max or abs only the finite values and leave the `NAN` values where they were in the original dataset.

Being new to ML, I wasn't sure if this was a valid approach for using the normalizers together with the Gower imputer - feedback welcome!

### DIFF
--- a/src/Transformers/MaxAbsoluteScaler.php
+++ b/src/Transformers/MaxAbsoluteScaler.php
@@ -100,7 +100,7 @@ class MaxAbsoluteScaler implements Transformer, Stateful, Elastic, Reversible, P
         foreach ($this->maxabs as $column => $oldMax) {
             $values = $dataset->feature($column);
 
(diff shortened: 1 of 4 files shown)


andrewdalpino:
Targetting ML 3.0 release with this since it can be construed as a backwards compatibility break.

PR #335, opened 2024-05-22
Do not allow zero local reachability density in LocalOutlierFactor

dmnc:
See https://github.com/RubixML/ML/issues/334

Sometimes we get an infinite k-distance which causes a zero local reachability density and a division by zero.


dmnc:

PR #337, opened 2024-05-26
Add a Median strategy

27pchrisl:
Hi,

This PR contributes the Median strategy for the Imputer.


andrewdalpino:
Hey thanks @27pchrisl, unfortunately we already have a Strategy that covers the median use-case that is the Percentile strategy. A 50th percentile is equivalent to the median.

https://github.com/RubixML/ML/blob/master/src/Strategies/Percentile.php

PR #338, opened 2024-05-27
Don't check for donors when fitting MissingDataImputer (use the strategy to check)

27pchrisl:
Hi,

During the fit stage of `MissingDataImputer`, it checks if there are donors: "Dataset must contain at least 1 donor per feature column.".

This is true for every strategy except the `Constant` strategy, where you don't need donors.
I checked, and every other strategy does its own check, throwing the right `Exception` if the provided array is empty.

This PR removes the check within `MissingDataImputer` so it's possible to use the `Constant` strategy with a dataset that has no donors.

### DIFF
--- a/src/Transformers/MissingDataImputer.php
+++ b/src/Transformers/MissingDataImputer.php
@@ -157,11 +157,6 @@ class MissingDataImputer implements Transformer, Stateful, Persistable
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


andrewdalpino:
Looks good to me, thank you for the great work @27pchrisl!

andrewdalpino (approved):

PR #346, opened 2024-10-29
Correct SVM Use

LouisAUTHIE:
Correcting the use of SVM for one class anomaly detection

### DIFF
--- a/src/AnomalyDetectors/OneClassSVM.php
+++ b/src/AnomalyDetectors/OneClassSVM.php
@@ -186,9 +186,8 @@ class OneClassSVM implements Estimator, Learner
         $data = [];
 
         foreach ($dataset->samples() as $sample) {
-            $sample_array = $sample;
-            array_unshift($sample_array, 1);
-            $data[] = $sample_array;
+            array_unshift($sample, 1);
+            $data[] = $sample;
         }
 
         $this->model = $this->svm->train($data);
--- a/src/AnomalyDetectors/OneClassSVM.php
+++ b/src/AnomalyDetectors/OneClassSVM.php
@@ -185,8 +185,10 @@ class OneClassSVM implements Estimator, Learner
 
         $data = [];
 
-        foreach ($dataset->samples() as $i => $sample) {
-            $data[] = array_merge([1], $sample);
+        foreach ($dataset->samples() as $sample) {
+            $sample_array = $sample;
+            array_unshift($sample_array, 1);
+            $data[] = $sample_array;
         }
 
         $this->model = $this->svm->train($data);
--- a/src/AnomalyDetectors/OneClassSVM.php
+++ b/src/AnomalyDetectors/OneClassSVM.php
@@ -79,6 +79,7 @@ class OneClassSVM implements Estimator, Learner
             new ExtensionIsLoaded('svm'),
             new ExtensionMinimumVersion('svm', '0.2.0'),
         ])->check();
+        
 
         if ($nu < 0.0 or $nu > 1.0) {
             throw new InvalidArgumentException('Nu must be between'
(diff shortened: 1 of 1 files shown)


src/AnomalyDetectors/OneClassSVM.php
```diff
@@ -182,7 +183,13 @@ public function train(Dataset $dataset) : void
             new SamplesAreCompatibleWithEstimator($dataset, $this),
         ])->check();
 
-        $this->model = $this->svm->train($dataset->samples());
+        $data = [];
+
+        foreach ($dataset->samples() as $i => $sample) {
```
andrewdalpino: Looks like we don't need this `$i` variable

src/AnomalyDetectors/OneClassSVM.php
```diff
@@ -182,7 +183,13 @@ public function train(Dataset $dataset) : void
...
+            $data[] = array_merge([1], $sample);
```
andrewdalpino: I wonder which is faster ...

```php
array_merge([1], $sample)
```

or

```php
array_unshift(1, $sample)
```

From what I recall, unshift is linear because it needs to reindex the array or something. I think merge is also linear.

src/AnomalyDetectors/OneClassSVM.php
```diff
@@ -211,7 +218,13 @@ public function predictSample(array $sample) : int
...
         }
 
-        return $this->model->predict($sample) !== 1.0 ? 0 : 1;
+        $sampleWithOffset = [];
+
+        foreach ($sample as $key => $value) {
+            $sampleWithOffset[$key + 1] = $value;
+        }
+
+        return $this->model->predict($sampleWithOffset) == 1 ? 0 : 1;
```
andrewdalpino, line 228: I notice we are "inversing" the logic here i.e. 1 is now 0, 0 is now 1. Is that intentional?

andrewdalpino (commented):
Nice @LouisAUTHIE thanks great work

I left a few comment and questions

src/AnomalyDetectors/OneClassSVM.php
LouisAUTHIE: Yes, in fact in the one class mode of libsvm, the "normal" samples are to be labelled with the 1 class. And the anomalies, are to be labelled with -1. That's why !

src/AnomalyDetectors/OneClassSVM.php
LouisAUTHIE: Change is done ;)

src/AnomalyDetectors/OneClassSVM.php
LouisAUTHIE: You are completely right, this is now corrected !

src/AnomalyDetectors/OneClassSVM.php
```diff
@@ -185,8 +185,10 @@ public function train(Dataset $dataset) : void
 
         $data = [];
 
-        foreach ($dataset->samples() as $i => $sample) {
-            $data[] = array_merge([1], $sample);
+        foreach ($dataset->samples() as $sample) {
```
andrewdalpino, line 188: So did array_unshift() turn out to be faster?

Is it necessary to assign the sample to an intermediate variable or would

```php
foreach ($dataset->samples() as $sample) {
    array_unshift($sample, 1);

    $data[] = $sample;
}
```

work here as well?

LouisAUTHIE: It is faster yes. And the rest works also, please see the next commit

andrewdalpino (approved):
Looks awesome, thank you @LouisAUTHIE

We'll get this deployed in a bugfix release ASAP

PR #351, opened 2025-01-31
Fix unnecessary logger in serialization

ElGigi:
Logger propery is in serialize data.
The fix remove property 'logger'.


andrewdalpino (approved):

andrewdalpino:
Looks good, thank you @ElGigi

PR #356, opened 2025-03-25
Convert initializers to NumPower

SkibidiProduction:
Initializers have been converted to NumPower.

For He, Xavier, LeCun initializers, a division into two types was made. The first type is based on a truncated normal distribution, the second on a unified distribution.

The final distributions of the initializers are made in accordance with their analogs in the keras library.

A fork of the original NumPower is used (https://github.com/RubixML/numpower)  due to the fact that there are difficulties with accepting changes and fixing bugs in the original extension.

### DIFF
--- a/src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
+++ b/src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rubix\ML\NeuralNet\Initializers\Base\Contracts;
+
+use Rubix\ML\NeuralNet\Initializers\Base\Exceptions\InvalidFanInException;
+use Rubix\ML\NeuralNet\Initializers\Base\Exceptions\InvalidFanOutException;
+
+/**
+ * Abstract Initializer for init params validation
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Andrew DalPino
+ * @author      Aleksei Nechaev <omfg.rus@gmail.com>
+ */
+abstract class AbstractInitializer implements Initializer
+{
+    /**
+     * Validating initializer parameters
+     *
+     * @param int $fanIn The number of input connections per neuron
+     * @param int $fanOut The number of output connections per neuron
+     * @throws InvalidFanInException Initializer parameter fanIn is less than 1
+     * @throws InvalidFanOutException Initializer parameter fanOut is less than 1
+     */
+    protected function validateInitParams(int $fanIn, int $fanOut) : void
+    {
+        if ($fanIn < 1) {
+            throw new InvalidFanInException(message: "Fan in cannot be less than 1, $fanIn given");
+        }
+
+        if ($fanOut < 1) {
+            throw new InvalidFanOutException(message: "Fan out cannot be less than 1, $fanOut given");
+        }
+    }
+}
--- a/src/NeuralNet/Initializers/Base/Contracts/Initializer.php
+++ b/src/NeuralNet/Initializers/Base/Contracts/Initializer.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Rubix\ML\NeuralNet\Initializers\Base\Contracts;
 
-use NumPower;
+use NDArray;
 use Stringable;
 use Rubix\ML\NeuralNet\Initializers\Base\Exceptions\InvalidFanInException;
 use Rubix\ML\NeuralNet\Initializers\Base\Exceptions\InvalidFanOutException;
@@ -26,7 +26,7 @@ interface Initializer extends Stringable
      * @param int<1, max> $fanOut The number of output connections per neuron
      * @throws InvalidFanInException Initializer parameter `fanIn` is less than 1
      * @throws InvalidFanOutException Initializer parameter `fanOut` is less than 1
-     * @return NumPower The initialized weight matrix of shape [fanOut, fanIn]
+     * @return NDArray The initialized weight matrix of shape [fanOut, fanIn]
      */
-    public function initialize(int $fanIn, int $fanOut) : NumPower;
+    public function initialize(int $fanIn, int $fanOut) : NDArray;
 }
--- a/src/NeuralNet/Initializers/Base/Contracts/Initializer.php
+++ b/src/NeuralNet/Initializers/Base/Contracts/Initializer.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rubix\ML\NeuralNet\Initializers\Base\Contracts;
+
+use NumPower;
+use Stringable;
+use Rubix\ML\NeuralNet\Initializers\Base\Exceptions\InvalidFanInException;
+use Rubix\ML\NeuralNet\Initializers\Base\Exceptions\InvalidFanOutException;
+
+/**
+ * Initializer
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Andrew DalPino
+ * @author      Aleksei Nechaev <omfg.rus@gmail.com>
+ */
+interface Initializer extends Stringable
+{
+    /**
+     * Initialize a weight matrix W in the dimensions `fanIn` x `fanOut`.
+     *
+     * @param int<1, max> $fanIn The number of input connections per neuron
+     * @param int<1, max> $fanOut The number of output connections per neuron
+     * @throws InvalidFanInException Initializer parameter `fanIn` is less than 1
+     * @throws InvalidFanOutException Initializer parameter `fanOut` is less than 1
+     * @return NumPower The initialized weight matrix of shape [fanOut, fanIn]
+     */
+    public function initialize(int $fanIn, int $fanOut) : NumPower;
+}
--- a/src/NeuralNet/Initializers/Base/Exceptions/InvalidFanInException.php
+++ b/src/NeuralNet/Initializers/Base/Exceptions/InvalidFanInException.php
@@ -1,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rubix\ML\NeuralNet\Initializers\Base\Exceptions;
+
+use Rubix\ML\Exceptions\InvalidArgumentException;
+
+/**
+ * Invalid initializer parameter `fanIn`
+ */
+class InvalidFanInException extends InvalidArgumentException
+{
+}
--- a/src/NeuralNet/Initializers/He/HeNormal.php
+++ b/src/NeuralNet/Initializers/He/HeNormal.php
@@ -6,7 +6,7 @@ namespace Rubix\ML\NeuralNet\Initializers\He;
 
 use NumPower;
 use NDArray;
-use Rubix\ML\NeuralNet\Initializers\Base\Contracts\AbstractInitializer;
+use Rubix\ML\NeuralNet\Initializers\Base\AbstractInitializer;
 
 /**
  * He Normal
@@ -31,11 +31,11 @@ class HeNormal extends AbstractInitializer
      */
     public function initialize(int $fanIn, int $fanOut) : NDArray
     {
-        $this->validateInitParams(fanIn: $fanIn, fanOut: $fanOut);
+        $this->validateFanInFanOut(fanIn: $fanIn, fanOut: $fanOut);
 
-        $std = sqrt(2 / $fanOut);
+        $stdDev = sqrt(2 / $fanOut);
 
-        return NumPower::truncatedNormal(size: [$fanOut, $fanIn], loc: 0.0, scale: $std);
+        return NumPower::truncatedNormal(size: [$fanOut, $fanIn], loc: 0.0, scale: $stdDev);
     }
 
     /**
--- a/src/NeuralNet/Initializers/He/HeNormal.php
+++ b/src/NeuralNet/Initializers/He/HeNormal.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rubix\ML\NeuralNet\Initializers\He;
+
+use NumPower;
+use Rubix\ML\NeuralNet\Initializers\Base\Contracts\AbstractInitializer;
+
+/**
+ * He Normal
+ *
+ * The He initializer was designed for hidden layers that feed into rectified
+ * linear layers such ReLU, Leaky ReLU, ELU, and SELU. It draws from a truncated
+ * normal distribution with mean 0 and standart deviation sqrt(2 / fanOut).
+ *
+ * References:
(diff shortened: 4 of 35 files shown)


src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
```
andrewdalpino: Is this necessary? If not let's not use strict types as it can impact runtime performance.

> Enabling strict types can improve code robustness and type safety, but it may also introduce performance overhead due to the additional type-checking at runtime.

src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Rubix\ML\NeuralNet\Initializers\Base\Contracts;
```
andrewdalpino: It would be fine to put this Abstract class directly in the Initializers namespace, without adding a `Base`  or introducing the concept of a "Contract." Rubix ML uses the term "interface" instead of contract and I wouldn't want to confuse people unnecessarily by introducing another name for something that already exists. In addition, this class does not actually provide an interface or contract, it's just a base class.

I think it makes sense to put Exception classes in their own folder, however.

How about we have

src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param int $fanIn The number of input connections per neuron
+     * @param int $fanOut The number of output connections per neuron
+     * @throws InvalidFanInException Initializer parameter fanIn is less than 1
+     * @throws InvalidFanOutException Initializer parameter fanOut is less than 1
+     */
+    protected function validateInitParams(int $fanIn, int $fanOut) : void
+    {
+        if ($fanIn < 1) {
+            throw new InvalidFanInException(message: "Fan in cannot be less than 1, $fanIn given");
```
andrewdalpino: To me, this exception does not meet the threshold for having it's own class. I would roughly define that threshold as being ...

1. Will the user want to catch this exception specifically? - IMO probably not
2. Does the exception have complex logic? - It does not

src/NeuralNet/Initializers/Base/Contracts/Initializer.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
andrewdalpino: Let's colocate the interface with their implementations and avoid creating unnecessary/duplicate semantics around base classes and interfaces. The PHP language already provides ubiquitous semantics for these concepts.

src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Validating initializer parameters
+     *
+     * @param int $fanIn The number of input connections per neuron
+     * @param int $fanOut The number of output connections per neuron
+     * @throws InvalidFanInException Initializer parameter fanIn is less than 1
+     * @throws InvalidFanOutException Initializer parameter fanOut is less than 1
+     */
+    protected function validateInitParams(int $fanIn, int $fanOut) : void
```
andrewdalpino: This method name is necessarily generic for what it's actually doing. Since this method specifically validates the fan in and fan out perhaps a more accurate name would be `validateFanInFanOut()`.

src/NeuralNet/Initializers/Base/Exceptions/InvalidFanInException.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
andrewdalpino, line 1: See my previous comment, but to reiterate here, I don''t think this exception warrants it's own class. I wouldn't want to introduce it unless there was a clear benefit. Is there something I may be missing?

src/NeuralNet/Initializers/He/HeNormal.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function initialize(int $fanIn, int $fanOut) : NDArray
+    {
+        $this->validateInitParams(fanIn: $fanIn, fanOut: $fanOut);
+
+        $std = sqrt(2 / $fanOut);
+
+        return NumPower::truncatedNormal(size: [$fanOut, $fanIn], loc: 0.0, scale: $std);
```
andrewdalpino: I'm assuming `loc` is referring to the expected mean of the distribution? What does `loc` stand for? Location? Lines of code?

src/NeuralNet/Initializers/Normal/Exceptions/InvalidStandardDeviationException.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
andrewdalpino, line 1: I have similar thoughts about this exception class as I do with the invalid fan in/fan out ones. I'm not seeing any extra logic and I don't foresee that users will need to catch this exception specifically.

src/NeuralNet/Initializers/Normal/Normal.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Return the string representation of the initializer.
+     *
+     * @return string String representation
+     */
+    public function __toString() : string
+    {
+        return "Normal (std: {$this->stdDev})";
```
andrewdalpino: For consistency, let's keep the string representation and the constructor argument names in sync. 

Ex.

```php
return "Normal (stdDev: {$this->stdDev})";
```

src/NeuralNet/Initializers/Normal/TruncatedNormal.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Return the string representation of the initializer.
+     *
+     * @return string String representation
+     */
+    public function __toString() : string
+    {
+        return "Truncated Normal (std: {$this->stdDev})";
```
andrewdalpino: Similar issue here with the string representation of the constructor argument being different than the name of the argument.

src/NeuralNet/Initializers/Uniform/Exceptions/InvalidBetaException.php
```diff
@@ -0,0 +1,14 @@
...
+declare(strict_types=1);
+
+namespace Rubix\ML\NeuralNet\Initializers\Uniform\Exceptions;
+
+use Rubix\ML\Exceptions\InvalidArgumentException;
+
+/**
+ * Invalid Uniform initializer parameter `beta`
+ */
+class InvalidBetaException extends InvalidArgumentException
```
andrewdalpino, line 12: I don't think we need this exception class.

src/NeuralNet/Initializers/Xavier/XavierNormal.php
```diff
@@ -0,0 +1,51 @@
...
+class XavierNormal extends AbstractInitializer
+{
+    /**
+     * @inheritdoc
+     */
+    public function initialize(int $fanIn, int $fanOut) : NDArray
+    {
+        $this->validateInitParams(fanIn: $fanIn, fanOut: $fanOut);
+
+        $std = sqrt(2 / ($fanOut + $fanIn));
```
andrewdalpino: For consistency, let's use the variable name `$stdDev`, I believe this is how we call it throughout the codebase.

tests/NeuralNet/Initializers/Constant/ConstantTest.php
```diff
@@ -0,0 +1,147 @@
+<?php
```
andrewdalpino, line 1: Nice thorough tests!

tests/NeuralNet/Initializers/Constant/ConstantTest.php
```diff
@@ -0,0 +1,147 @@
...
+            ],
+            'fanIn and fanOut less than 1' => [
+                'fanIn' => 0,
+                'fanOut' => 0,
+            ],
+        ];
+    }
+
+    #[Test]
+    #[TestDox('The initializer object os created correctly')]
```
andrewdalpino: Typo

```php
#[TestDox('The initializer object is created correctly')]
```

tests/NeuralNet/Initializers/Constant/ConstantTest.php
```diff
@@ -0,0 +1,147 @@
...
+        $this->expectNotToPerformAssertions();
+
+        //when
+        new Constant($value);
+    }
+
+    #[Test]
+    #[TestDox('The result matrix has correct shape')]
+    #[DataProvider('initializeTest1DataProvider')]
+    public function initializeTest1(int $fanIn, int $fanOut) : void
```
andrewdalpino: Can we make these test method names more descriptive?

Example `initializeWithCorrectShape()`

tests/NeuralNet/Initializers/Constant/ConstantTest.php
```diff
@@ -0,0 +1,147 @@
...
+        $values = $w->toArray();
+
+        //then
+        $this->assertEquals([4.5], array_unique(array_merge(...$values)));
+    }
+
+    #[Test]
+    #[TestDox('An exception is thrown during initialization')]
+    #[DataProvider('initializeTest3DataProvider')]
+    public function initializeTest3(int $fanIn, int $fanOut) : void
```
andrewdalpino: Can we make these test method names more descriptive?

Example `initializeWithBadFanInFanOut()`

andrewdalpino (commented):
Hey @SkibidiProduction overall a pretty good start!

I really like the thoroughness of the unit tests but would encourage you to come up with some more descriptive names for the test methods instead of `initializeTest1()` `initializeTest2()` etc. I also like that you broke the initializers into Normal and Uniform versions. Finally, I like that you converted to more modern PHP syntax.

I'm concerned with strict types being unnecessary and slowing down our runtime. I'm also not keen on the idea of introducing additional exception classes without a compelling reason to do so. Also, I think we should stick to calling classes that provide API as "interfaces" and not introduce the "contract" term to mean the same thing as it has the potential to confuse people but provides no additional benefit. Lastly, I think we should avoid creating additional namespace hierarchies unless they are necessary.

src/NeuralNet/Initializers/He/HeNormal.php
SkibidiProduction: Naming same as in numpy https://numpy.org/doc/2.1/reference/random/generated/numpy.random.normal.html

src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
SkibidiProduction: We agreed to leave it at that

src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
SkibidiProduction: We agreed to leave it temporary

src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
andrewdalpino: Yes, thank you for running those benchmarks too.

marclaporte:
> A fork of the original NumPower is used (https://github.com/RubixML/numpower) due to the fact that there are difficulties with accepting changes and fixing bugs in the original extension.

Once things are tested, would it make sense to propose changes as a PR at https://github.com/NumPower/numpower/pulls ?

So these changes would have more visibility and perhaps lead to more collaboration throughout the ecosystem.

SkibidiProduction:
> Once things are tested, would it make sense to propose changes as a PR at https://github.com/NumPower/numpower/pulls ?
> 
> So these changes would have more visibility and perhaps lead to more collaboration throughout the ecosystem.

@marclaporte Yes, but it's better not to change old classes, but to create new ones. Otherwise, tests will fail and this will complicate refactoring.

src/NeuralNet/Initializers/Base/Contracts/Initializer.php
SkibidiProduction: The Contracts namespace has been removed

src/NeuralNet/Initializers/Base/Contracts/AbstractInitializer.php
SkibidiProduction: the name has been changed

src/NeuralNet/Initializers/Base/Exceptions/InvalidFanInException.php
SkibidiProduction: We discussed that we would keep this approach to reduce testing entropy.

src/NeuralNet/Initializers/Normal/Normal.php
SkibidiProduction: The name has been changed

tests/NeuralNet/Initializers/Constant/ConstantTest.php
SkibidiProduction: fixed

tests/NeuralNet/Initializers/Constant/ConstantTest.php
SkibidiProduction: More meaningful names have been selected for the methods.

andrewdalpino:
Which tests will fail @SkibidiProduction? Are we able to modify the tests along with the changes to the classes so they do not fail?

> > Once things are tested, would it make sense to propose changes as a PR at https://github.com/NumPower/numpower/pulls ?
> > So these changes would have more visibility and perhaps lead to more collaboration throughout the ecosystem.
> 
> @marclaporte Yes, but it's better not to change old classes, but to create new ones. Otherwise, tests will fail and this will complicate refactoring.

andrewdalpino:
Looks good, nice work @SkibidiProduction

andrewdalpino (approved):

SkibidiProduction:
> Which tests will fail @SkibidiProduction? Are we able to modify the tests along with the changes to the classes so they do not fail?

@andrewdalpino Initializers now return NDArray instead of Matrix as before. However, other entities, such as layers, optimizers, etc., cannot work with NDArray objects. Moreover, they use Matrix objects as input parameters, and use Matrix objects inside their implementation. Thus, changing the entity of one type used in building a neural network, or even more so changing only one class, will lead to the tests for all related code falling.

andrewdalpino:
Oh yes @SkibidiProduction I know that, I meant that question in the context of NumPower. @marclaporte suggested we submit our changes to the original NumPower repo when we're done making changes needed for Rubix ML. I think that's a worthwhile thing to do, but it sounded like you suggested there would be issues, specifically with tests.

PR #367, opened 2025-06-17
Sam 2 refactor elu function

apphp:


src/NeuralNet/FeedForward.php
```diff
@@ -27,8 +27,7 @@
  * @package     Rubix/ML
  * @author      Andrew DalPino
  */
-class FeedForward implements Network
```
apphp, line 30: There was a mistake here, because Network is no longer an Interface, it is now a regular class.

PR #381, opened 2025-07-14
Convert activation functions to numpower

SkibidiProduction:
**Activation implementations**

- Swapped out custom Tensor code for NumPower APIs across all functions: ReLU, LeakyReLU, ELU, GELU, HardSigmoid, SiLU, Tanh, Sigmoid, Softmax, Softplus, Softsign, ThresholdedReLU, etc.

- Updated derivative methods to use numpower’s derivative helpers.

**Tests**

- Refactored unit tests to assert against numpower outputs.

- Adjusted tolerances and assertions to match numpower’s numeric behavior.

**Documentation**

- Added/updated images under docs/images/activation-functions/ to illustrate each activation curve and its derivative using the new implementations.

- Cleaned up corresponding markdown to reference the updated diagrams.

**Code cleanup**

- Aligned naming conventions and method signatures with numpower’s API.

- Minor style fixes (whitespace, imports, visibility).


.github/workflows/ci.yml
```diff
@@ -8,14 +8,8 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
-<<<<<<< HEAD
```
andrewdalpino, line 11: Hahaha how did this get there?

.php-cs-fixer.dist.php
```diff
@@ -3,115 +3,128 @@
 use PhpCsFixer\Finder;
 use PhpCsFixer\Config;
 
-$finder = Finder::create()->in(__DIR__)
-    ->exclude('docs');
+$finder = Finder::create()
+    ->exclude([
```
andrewdalpino, line 7: I am curious what is the difference with this new configuration?

docs/images/activation-functions/elu-derivative.png
andrewdalpino, line 1: Love love love these chart visuals!

Just to be sure, for copyright purposes, were these generated by you? If not, are they permitted to be copied?

docs/neural-network/activation-functions/elu.md
```diff
@@ -1,23 +1,35 @@
-<span style="float:right;"><a href="https://github.com/RubixML/ML/blob/master/src/NeuralNet/ActivationFunctions/ELU.php">[source]</a></span>
+<span style="float:right;"><a href="https://github.com/RubixML/ML/blob/master/src/NeuralNet/ActivationFunctions/ELU/ELU.php">[source]</a></span>
 
 # ELU
 *Exponential Linear Units* are a type of rectifier that soften the transition from non-activated to activated using the exponential function. As such, ELU produces smoother gradients than the piecewise linear [ReLU](relu.md) function.
 
 $$
-{\displaystyle ELU = {\begin{cases}\alpha \left(e^{x}-1\right)&{\text{if }}x\leq 0\\x&{\text{if }}x>0\end{cases}}}
+\text{ELU}(x) =
```
andrewdalpino, line 7: Very nice, are you familiar with this math syntax? When I wrote these docs, I was not nor am I still.

Was there an error in the original notation or did something change with the implementation?

docs/neural-network/activation-functions/elu.md
```diff
@@ -1,23 +1,35 @@
...
+\end{cases}
 $$
 
 ## Parameters
 | # | Name | Default | Type | Description |
 |---|---|---|---|---|
 | 1 | alpha | 1.0 | float | The value at which leakage will begin to saturate. Ex. alpha = 1.0 means that the output will never be less than -1.0 when inactivated. |
 
+## Size and Performance
+ELU is a simple function and is well-suited for deployment on resource-constrained devices or when working with large neural networks.
```
andrewdalpino: How did you come up with these size and performance details? I'm noticing that some differ from my understanding. For example, it is not necessarily true when taken in the context of all activation functions that ELU is a simple function or well-suited for resource constrained devices.

Perhaps it would actually be more confusing to offer this somewhat subjective explanation. In addition, in practice, activation functions have very little impact on the total runtime of the network - so taking the effort here to detail out their performance is somewhat distracting.

How do you feel about dropping this "size and performance" section all together, not being opinionated about individual activation functions, and instead letting the user discover the nuances of each activation function for themselves? However, if there is something truly outstanding about a particular activation functions performance characteristics, then let's make sure to include that in the description of the class. For example, ReLU is outstanding because it is the simplest activation function in the group. Maybe there's another activation function that has an associated kernel that is particularly optimized, etc.

docs/transformers/regex-filter.md
```diff
@@ -28,7 +28,6 @@ $transformer = new RegexFilter([
 ```
 
 ## Predefined Regex Patterns
-<<<<<<< HEAD
```
andrewdalpino, line 31: More?

src/AnomalyDetectors/OneClassSVM.php
```diff
@@ -85,7 +85,7 @@ public function __construct(
                 . "0 and 1, $nu given.");
         }
 
-        $kernel = $kernel ?? new RBF();
+        $kernel ??= new RBF();
```
andrewdalpino, line 88: Love this

src/NeuralNet/ActivationFunctions/Base/Contracts/IBufferDerivative.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Derivative based on input buffer
+ *
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Andrew DalPino
+ * @author      Aleksei Nechaev <omfg.rus@gmail.com>
+ * @author      Samuel Akopyan <leumas.a@gmail.com>
+ */
+interface IBufferDerivative extends Derivative
```
andrewdalpino, line 18: Just so I understand 100%, we're breaking out these interfaces to account for the fact that some activation functions will memorize and use the function input in computing the derivative, some will memorize and use the output, and some will need both - is that correct?

src/NeuralNet/ActivationFunctions/HardSigmoid/HardSigmoid.php
```diff
@@ -0,0 +1,109 @@
...
+ * @author      Samuel Akopyan <leumas.a@gmail.com>
+ */
+class HardSigmoid implements ActivationFunction, IBufferDerivative
+{
+    /**
+     * The slope of the linear region.
+     *
+     * @var float
+     */
+    protected const SLOPE = 0.2;
```
andrewdalpino, line 33: Fantastic usage of class constants and nicely done using it as an opportunity to document the reason why the constant is special.

src/NeuralNet/ActivationFunctions/Softmax/Softmax.php
```diff
@@ -0,0 +1,101 @@
...
+     * each activation between 0 and 1, and all activations add up to 1.
+     *
+     * > **Note:** This function can be rewritten in a more efficient way,
+     * using NumPower::exp(), NumPower::sum(), and NumPower::divide().
+     * Currently blocked by implementation of 2nd parameter "axis" for NumPower::sum()
+     *
+     * @param NDArray $input
+     * @return NDArray
+     */
+    public function activate(NDArray $input) : NDArray
```
andrewdalpino, line 41: For softmax, perhaps we should implement a `softmax()` function in NumPower if there isn't one already.

For some architectures, doing the softmax in PHPland won't be so noticeable, but in some it will be very noticable.

src/NeuralNet/ActivationFunctions/Softmax/Softmax.php
```diff
@@ -0,0 +1,101 @@
...
+
+        // Process each row separately to ensure row-wise normalization
+        foreach ($inputArray as $row) {
+            $expRow = array_map('exp', $row);
+            $sum = array_sum($expRow);
+            $softmaxRow = [];
+
+            foreach ($expRow as $value) {
+                // Round to 7 decimal places to match test expectations
+                $softmaxRow[] = round($value / $sum, 7);
```
andrewdalpino, line 55: Why does the test expect only 7 decimal places? I would expect the test to expect the unrounded value with a small delta to account for platform differences.

src/NeuralNet/ActivationFunctions/Softsign/Softsign.php
```diff
@@ -0,0 +1,81 @@
...
+     * Compute the activation.
+     *
+     * f(x) = x / (1 + |x|)
+     *
+     * @param NDArray $input
+     * @return NDArray
+     */
+    public function activate(NDArray $input) : NDArray
+    {
+        // Calculate |x|
```
andrewdalpino: I don't feel that these comments provide enough value to justify their existence. I can understand what is going on clearly given your great usage of variables and naming.

tests/NeuralNet/ActivationFunctions/ELU/ELUTest.php
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    #[Test]
+    #[TestDox('Correctly activates the input')]
+    #[DataProvider('computeProvider')]
+    public function testActivate(NDArray $input, array $expected) : void
+    {
+        $activations = $this->activationFn->activate($input)->toArray();
+
+        static::assertEqualsWithDelta($expected, $activations, 1e-7);
```
andrewdalpino, line 127: Related to the comment about rounding to the 7th decimal place (which I don't think we should be doing) ... The 1e-7 delta here is to account for platform differences (32 vs 64 bit for example) and not to suggest that the value should be rounded to the 7th decimal place. In actuality the activation should remain in the same precision that it was computed in.

andrewdalpino (commented):
Very nice work @apphp and @SkibidiProduction ... I think this is exactly what we need for the first round of integration with NumPower. I had a few questions and comments that may change the outcome of the PR so I'm just going to leave it at that for now until we get that sorted.

Overall, fantastic usage of unit tests and good code quality. I love to see it.

Andrew

docs/images/activation-functions/elu-derivative.png
apphp: I created them by running python code (using pyplot) for every function.

src/NeuralNet/ActivationFunctions/Softmax/Softmax.php
SkibidiProduction: Unless otherwise stated, NumPower uses float32 data type. However, PHP uses float64, and conversion on input may result in loss of precision.

Probably when I finish my work on types we will use float64 by default. But not now.

tests/NeuralNet/ActivationFunctions/ELU/ELUTest.php
SkibidiProduction: In the current version of NumPower API, only float32 is available by default. I'm working on it.

src/NeuralNet/ActivationFunctions/Base/Contracts/IBufferDerivative.php
SkibidiProduction: Yes

src/NeuralNet/ActivationFunctions/Softmax/Softmax.php
SkibidiProduction: Yes, I will do it

.github/workflows/ci.yml
apphp: May be from one of previous merges

.php-cs-fixer.dist.php
apphp: Nothing, I just added exclusion for vendor/ directory and specify cache file location in runtime/ directory. Just make more order in cache files.

docs/neural-network/activation-functions/elu.md
apphp: Original is not correctly works on GitHub - GitHub not always correctly shows standard syntax of LaTeX.

See original here: https://github.com/RubixML/ML/blob/master/docs/neural-network/activation-functions/elu.md
image

apphp: - But now it looks as expected
See: https://github.com/RubixML/ML/blob/convert-activation-functions-to-numpower/docs/neural-network/activation-functions/elu.md
image

apphp (changes requested):

docs/neural-network/activation-functions/elu.md
apphp: Fixed in https://github.com/RubixML/ML/pull/382

src/NeuralNet/ActivationFunctions/Softsign/Softsign.php
apphp: Will be removed

apphp (changes requested):

.php-cs-fixer.dist.php
andrewdalpino: Awesome

docs/images/activation-functions/elu-derivative.png
andrewdalpino: Sweet, nice job

docs/neural-network/activation-functions/elu.md
andrewdalpino: Aha, glad it's fixed

docs/neural-network/activation-functions/elu.md
andrewdalpino: Yes remove the section, but if there is something unique about a particular functions performance characteristics, we can put that info in the description. What do you think?

src/NeuralNet/ActivationFunctions/Softmax/Softmax.php
andrewdalpino: Awesome, thanks for the context

docs/neural-network/activation-functions/elu.md
apphp: Will be removed

apphp: Fixed in https://github.com/RubixML/ML/pull/382

src/NeuralNet/ActivationFunctions/Softsign/Softsign.php
apphp: Fixed in https://github.com/RubixML/ML/pull/382

andrewdalpino (approved):
Looks good fellas, let's roll!

PR #382, opened 2025-08-06
Removed unneeded comment

apphp:


src/NeuralNet/ActivationFunctions/Softsign/Softsign.php
```diff
@@ -36,7 +36,6 @@ class Softsign implements ActivationFunction, IBufferDerivative
      */
     public function activate(NDArray $input) : NDArray
     {
-        // Calculate |x|
```
andrewdalpino, line 39: Nice, but what about the rest of the comments? I might have not been as clear as I could have in my review, sorry for that, but I meant it to apply to all of these types of comments in this class and any similar class not just the one on this line.

apphp: Done, please re-view

andrewdalpino (approved):
Looks good @apphp, thank you

PR #387, opened 2025-10-13
Ml 385 convert cost functions to num power

apphp:


src/NeuralNet/CostFunctions/Base/Contracts/CostFunction.php
```diff
@@ -0,0 +1,40 @@
...
+ * @category    Machine Learning
+ * @package     Rubix/ML
+ * @author      Samuel Akopyan <leumas.a@gmail.com>
+ */
+interface CostFunction extends Stringable
+{
+    /**
+     * Compute the loss score.
+     *
+     * @internal
```
SkibidiProduction: I think this could be removed so that the user would have the ability to create their own loss function based on this interface. Same below.

apphp: Done

src/NeuralNet/CostFunctions/CrossEntropy/CrossEntropy.php
```diff
@@ -0,0 +1,91 @@
...
+ * @author      Samuel Akopyan <leumas.a@gmail.com>
+ */
+class CrossEntropy implements ClassificationLoss
+{
+    use ValidatesShapes;
+
+    /**
+     * Compute the loss score.
+     *
+     * L(y, ŷ) = -Σ(y * log(ŷ)) / n
```
andrewdalpino, line 35: Love how you put the formula here!

src/NeuralNet/CostFunctions/CrossEntropy/CrossEntropy.php
```diff
@@ -0,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rubix\ML\NeuralNet\CostFunctions\CrossEntropy;
```
andrewdalpino, line 5: Are you are carving out this namespace to hold the new classes temporarily until we can get the whole Neural Net subsystem converted to NumPower?

src/Traits/ValidatesShapes.php
```diff
@@ -0,0 +1,35 @@
...
+trait ValidatesShapes
+{
+    /**
+     * Validate that output and target have the same shape.
+     *
+     * @param NDArray $output
+     * @param NDArray $target
+     * @throws InvalidArgumentException
+     */
+    protected function validateShapes(NDArray $output, NDArray $target) : void
```
andrewdalpino: To me, this function name does not describe what the function really does. I would recommend something like "validateShapesAreEqual()" or better yet, this should be a method on the NDArray object. Something to consider implementing in NumPower @SkibidiProduction.

andrewdalpino (commented):
Looks great @apphp! I left a couple comments one on function naming and the other just to clarify our temporary namespacing. Other than that, I think it's good to go. I really liked the additional context you provided by the comments ... I'm generally against comments unless they bring outsized value and you hit it just right with the addition of the mathematical formulas.

andrewdalpino (approved):

src/NeuralNet/CostFunctions/CrossEntropy/CrossEntropy.php
apphp: Yes, exactly - like it was done by @SkibidiProduction for activation functions.

src/Traits/ValidatesShapes.php
apphp: Ok, agree. Renamed into "assertSameShape"

apphp: When we add this method to NDArray object, then we can remove this trait

PR #389, opened 2025-11-04
388 convert parameter class to numpower

apphp:


andrewdalpino (approved):
Looks good to me! Nice work

PR #391, opened 2025-11-07
390 convert optimizers class to num power

apphp:


src/NeuralNet/Optimizers/Cyclical/Cyclical.php
```diff
@@ -0,0 +1,166 @@
...
+
+        $x = abs($this->t / $this->losses - 2 * $cycle + 1);
+
+        $scale = $this->decay ** $this->t;
+
+        $rate = $this->lower + $this->range * max(0, 1 - $x) * $scale;
+
+        ++$this->t;
+
+        return NumPower::multiply($gradient, $rate);
```
andrewdalpino, line 151: Is there an API to do `$gradient->multiply($rate)`?

apphp: No, all operations currently implemented only in NumPower class, not in NDArray

andrewdalpino (approved):
Looks good I just wish we could change the API of NumPower to look like

```php
$a->multiply($b);
```

instead of

```php
NumPower::multiple($a, $b);
```

PR #393, opened 2025-12-06
392 convert layers to numpower

apphp:


src/NeuralNet/Layers/BatchNorm/BatchNorm.php
```diff
@@ -0,0 +1,389 @@
...
+     * @var NDArray|null
+     */
+    protected ?NDArray $xHat = null;
+
+    /**
+     * Row-wise or column-wise normalization.
+     *
+     * @var int
+     */
+    protected const int AXIS_SAMPLES = 0;
```
andrewdalpino: I would just double check that this is correct and not the other way around. From my understanding the input to each layer is shaped [num_features, num_samples] which would mean that axis 0 is the features not the samples.

src/NeuralNet/Layers/BatchNorm/BatchNorm.php
```diff
@@ -0,0 +1,389 @@
...
+    public function gradient(NDArray $dOut, NDArray $gamma, NDArray $stdInv, NDArray $xHat) : NDArray
+    {
+        $dXHat = NumPower::multiply($dOut, $gamma);
+        $xHatSigma = NumPower::sum(NumPower::multiply($dXHat, $xHat), self::AXIS_FEATURES);
+        $dXHatSigma = NumPower::sum($dXHat, self::AXIS_FEATURES);
+
+        $m = $dOut->shape()[0];
+
+        // Compute gradient per formula: dX = (dXHat * m - dXHatSigma - xHat * xHatSigma) * (stdInv / m)
+        return NumPower::multiply(
```
andrewdalpino: Can we break this apart to help make it more human-readable?

src/NeuralNet/Layers/Binary/Binary.php
```diff
@@ -0,0 +1,222 @@
...
+     * @param NDArray $input
+     * @param NDArray $output
+     * @param NDArray $expected
+     * @return NDArray
+     */
+    public function gradient(NDArray $input, NDArray $output, NDArray $expected) : NDArray
+    {
+        $n = $output->shape()[1];
+
+        if ($this->costFn instanceof CrossEntropy) {
```
andrewdalpino, line 193: I don't remember if I left a comment here or not in the original implementation, but it's worth noting to future developers that this is an optimization specific to the Binary Cross Entropy loss function. Through some clever mathematical cancellation tricks  this gradient calculation is just a simple subtraction operation.

src/NeuralNet/Layers/Dense/Dense.php
```diff
@@ -0,0 +1,348 @@
...
+    {
+        if (!$this->weights) {
+            throw new RuntimeException('Layer is not initialized');
+        }
+
+        $output = NumPower::matmul($this->weights->param(), $input);
+
+        if ($this->biases) {
+            // Reshape bias vector [fanOut] to column [fanOut, 1] to match output [fanOut, n]
+            $bias = NumPower::reshape($this->biases->param(), [$this->neurons, 1]);
```
andrewdalpino, line 197: Was this reshaping required in the original Tensor code?

src/NeuralNet/Layers/Swish/Swish.php
```diff
@@ -0,0 +1,303 @@
...
+            throw new RuntimeException('Must perform forward pass before backpropagating.');
+        }
+
+        /** @var NDArray $dOut */
+        $dOut = $prevGradient();
+
+        // Gradient of the loss with respect to beta
+        // dL/dbeta = sum_over_batch(dL/dy * dy/dbeta)
+        // Here we use a simplified formulation: dL/dbeta ~ sum(dOut * input)
+        $dBetaFull = NumPower::multiply($dOut, $this->input);
```
andrewdalpino, line 183: Interesting comment, you know I'm not entirely sure I had the original gradient calculation 100% correct in the first place.

andrewdalpino (commented):
This looks great @apphp, mostly good to go I just had a few questions before I approved.

src/NeuralNet/Layers/BatchNorm/BatchNorm.php
apphp: Done

src/NeuralNet/Layers/BatchNorm/BatchNorm.php
apphp: Done. It was rewritten to use the convention [num_features, num_samples], meaning features-first.

src/NeuralNet/Layers/Binary/Binary.php
apphp: No, but this is was a good point - I added comment for both implementations.

src/NeuralNet/Layers/Dense/Dense.php
apphp: No, original Tensor code didnt need it, but NumPower version does to get correct bias broadcasting across the batch dimension. 

This is because in the NumPower NumPower::add($output, $biasVector) does not reliably broadcast a 1-d vector across the second dimension

src/NeuralNet/Layers/Swish/Swish.php
apphp: Do you want me refactor this calculation?

andrewdalpino: Let's leave it for now as these change are more about converting rather than fixing. But if indeed the calculation is wrong then we'll fix it in a followup.

src/NeuralNet/Layers/Dense/Dense.php
andrewdalpino: That's great to know, thanks!

andrewdalpino:
At some point we should resolve the static analysis warnings as well

andrewdalpino (approved):

PR #399, opened 2026-02-22
1D Conv, Avg Pool, and Max Pool

Boorinio:
Adds ConvD1,Max/Avg layers :D


andrewdalpino (commented):
Hey this is great @Boorinio , before I dive into the review can I ask where you obtained the formula/calculation of the gradient for 1D Conv? Is there a reference?

It's great that you;ve also included pooling layers, I'm going to change the title of the PR to reflect the additional classes.

Boorinio:
> Hey this is great @Boorinio , before I dive into the review can I ask where you obtained the formula/calculation of the gradient for 1D Conv? Is there a reference?
> 
> It's great that you;ve also included pooling layers, I'm going to change the title of the PR to reflect the additional classes.

Added numerical checks for the layers as the review from copilot makes sense.
Links that were used for the tests and for the layers

https://www.michaelpiseno.com/pages/conv-backprop.html
https://maxkendall.com/blog_posts/cnn_backpropagation.html
https://anhreynolds.com/blogs/cnn.html

also light context read https://cs231n.stanford.edu/slides/2025/lecture_4.pdf :P

src/NeuralNet/Layers/AvgPool1D.php
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        $batchSize = (int) ($input->n() / $this->inputLength);
+
+        $inputArray = $input->asArray();
+        $output = [];
+
+        $scale = 1.0 / $this->poolSize;
+
+        foreach ($inputArray as $channel) {
```
andrewdalpino, line 171: Let's leverage the existing functionality in the underlying tensor library to compute the 1D convolution.

https://github.com/RubixML/Tensor/blob/master/src/Vector.php#L530

src/NeuralNet/Layers/MaxPool1D.php
```diff
@@ -0,0 +1,344 @@
...
+                . " expected {$this->inputChannels}, got {$inputChannels}.");
+        }
+
+        $batchSize = (int) ($input->n() / $this->inputLength);
+
+        $inputArray = $input->asArray();
+        $output = [];
+        $maxIndices = [];
+
+        foreach ($inputArray as $c => $channel) {
```
andrewdalpino, line 177: Again, let's use the underlying tensor library for the max() operation if we can.

andrewdalpino (changes requested):
Hi @Boorinio, I took a quick look and noticed that most if not all of the computation is being done using PHP array instead of Tensors. Let's use Tensor operations.

https://github.com/RubixML/Tensor

andrewdalpino:
Hey @Boorinio, I've been trying to think of some use-cases for these layers in the context of Rubix ML. Did you have any particular use-cases in mind? Reminder that we're limited to a tabular dataset in ML 2.0.

Boorinio:
> Hey @Boorinio, I've been trying to think of some use-cases for these layers in the context of Rubix ML. Did you have any particular use-cases in mind? Reminder that we're limited to a tabular dataset in ML 2.0.


Hey sorry for the late reply, yeah i currently use it in 2 places, one is to train a model to detect skin lesions (during training we randomly augment the images and we also add certain features like hair) and the second model tracks growth. Both models saw increased accuracy over this. 


One another note, would you accept prs to change that? maybe change the input to a shape class so in the future we can proceed with multidimensional inputs?

src/NeuralNet/Layers/AvgPool1D.php
Boorinio: Are we sure we want to do that? wouldn't it create a bit of overhead to do a tensor in the loop ??

andrewdalpino:
Hey @Boorinio, can you share your example in code? Are you using Rubix ML for this? If so, are you flattening the images and doing a 1D convolution on that?

We could create a new Dataset class to handle volumes for example. The issue beyond that is Tensor is not meant for n-dimensional tensors. Perhaps this is a good enhancement to consider for ML 3.1 since by that point we'll be using an ND array to power the neural network subsystem.

Boorinio:
> Hey @Boorinio, can you share your example in code? Are you using Rubix ML for this? If so, are you flattening the images and doing a 1D convolution on that?
> 
> 
> 
> We could create a new Dataset class to handle volumes for example. The issue beyond that is Tensor is not meant for n-dimensional tensors. Perhaps this is a good enhancement to consider for ML 3.1 since by that point we'll be using an ND array to power the neural network subsystem.

Yeah it is Rubix, it used to be a tensorflow model but it actually helped us save some costs and complexity migrating that and the accuracy was pretty good, unfortunately I can't share the code in a public setting 😅. Do you have a projection on when 3.1 would be released?

andrewdalpino:
Hi @Boorinio now it's my turn to reply late. I feel like we'll have 3.0 out by the end of the year. At that point we'll start working on 3.1.

PR #402, opened 2026-03-10
400 refactor network class to interface

apphp:


apphp:
All tests passed
image

PR #403, opened 2026-03-11
ML-400 Converted back Network interface

apphp:


apphp:
All tests were passed

image

src/NeuralNet/Network.php
```diff
@@ -2,270 +2,24 @@
...
-                $dot .= "  N{$parentId} -> N{$layerNum};" . PHP_EOL;
-            }
-        }
-
-        $dot .= '}';
-
-        return new Encoding($dot);
-    }
+    public function layers() : Traversable;
 }
```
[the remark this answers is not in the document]
apphp: Fixed

CHANGELOG.md
```diff
@@ -8,7 +8,9 @@
     - Removed output layer L2 Penalty parameter from MLP Learners
     - Remove Network interface
     - RBX Serializer only tracks major library version number
-    
+    - Convert NeuralNet classes to use NDArray instead of Matrix
+    - Converted back Network interface
```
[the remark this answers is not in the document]
apphp: Fixed

src/NeuralNet/Networks/FeedForward/FeedForward.php
```diff
@@ -273,4 +300,16 @@ public function exportGraphviz() : Encoding
...
     }
+
+    /**
+     * Normalize samples to a strict list-of-lists with sequential numeric keys.
+     * NumPower's C extension expects packed arrays and can error or behave unpredictably
+     * when given arrays with non-sequential keys (e.g. after randomize/take/fold operations).
+     * @param array $samples
+     * @return array
+     */
+    private function normalizeSamples(array $samples) : array
```
andrewdalpino: Let's avoid using the word "normalize" here as it's already used in the library to mean different things and this departs from the typical ML usage of the word.

Instead I would recommend something like `packSamples()`, or `arrayToPackedArray()`.

In addition, since this is just a general function it does not need to be a method. If you prefer to make it a member of the class make it a static member instead otherwise consider implementing this as a helper function.

https://github.com/RubixML/ML/blob/master/src/functions.php

src/NeuralNet/Networks/FeedForward/FeedForward.php
```diff
@@ -90,6 +96,8 @@ public function __construct(Input $input, array $hidden, Output $output, Optimiz
         $this->output = $output;
         $this->optimizer = $optimizer;
         $this->backPass = $backPass;
+
+        $this->normalizeSamples = false;
```
andrewdalpino: This looks like a constant. If so, remove it including the conditional logic in the infer() method. If not, provide an API for changing its value.

Why do we not always pack the samples array?

andrewdalpino (commented):
Left some comments for you @apphp

src/NeuralNet/Networks/FeedForward/FeedForward.php
apphp: done

src/NeuralNet/Networks/FeedForward/FeedForward.php
apphp: 1. Done
2. Generally it should work with NDArray, but for some reasons sometime it doesn't. I asked Alesei Nechav to check this. We don't want to use it always, because this operation requires additional time (small, but sill an additional time). Let's leave this currently as an option.

src/NeuralNet/Networks/FeedForward/FeedForward.php
andrewdalpino: Nice, if you didn't want to make that pack array helper function in to a global function (it likely is a global function as it applies to all PHP arrays) then at least make it a static member of the class it's implemented in - as it is definitely not a method.

src/NeuralNet/Networks/FeedForward/FeedForward.php
andrewdalpino: What is the runtime overhead of packing a matrix? Let's use a 1M element square matrix as the baseline.

src/NeuralNet/Networks/FeedForward/FeedForward.php
apphp: Sure, extracted to global reindex_nested_array() method and covered with unit tests.

src/NeuralNet/Networks/FeedForward/FeedForward.php
apphp: Tested on PHP8.4 
- 1M element square matrix takes 0.001 sec.
- 4M element square matrix takes 0.008 sec.
- 9M element square matrix takes 0.025 sec. but only if increase memory_limit to 512M

andrewdalpino (approved):
Looks good!

PR #415, opened 2026-07-06
Add check for NumPower at runtime instead of installtime

andrewdalpino:
@SkibidiProduction should we check for a minimum version as well?


src/NeuralNet/ActivationFunctions/ELU.php
```diff
@@ -41,6 +44,11 @@ public function __construct(protected float $alpha = 1.0)
                 message: "Alpha must be greater than 0, {$this->alpha} given."
             );
         }
+
+        SpecificationChain::with([
+            new ExtensionIsLoaded('RubixNumPower'),
+            new ExtensionMinimumVersion('RubixNumPower', '0.7.0'),
```
apphp, line 50: We have this in many places:
 
new ExtensionMinimumVersion('svm', '0.2.0'),
new ExtensionMinimumVersion('RubixNumPower', '0.7.0'),

what do you think, can create global consts for these version or small config file with min.requirements?

apphp (changes requested):

src/NeuralNet/ActivationFunctions/ELU.php
andrewdalpino: Hmmmmmm let's give that some thought.

For now it works, but it may not work in the future if/when different classes require different minimum versions of NumPower.

apphp:
@andrewdalpino we have some conflicts after my branch was merged, please check to resolve them.

apphp (approved):

PR #416, opened 2026-07-07
396 convert regressors to num power 1

apphp:


src/Regressors/MLPRegressor.php
```diff
@@ -233,6 +241,7 @@ public function __construct(
         $this->holdOut = $holdOut;
         $this->costFn = $costFn ?? new LeastSquares();
         $this->metric = $metric ?? new RMSE();
+        $this->packSamples = $packSamples;
```
andrewdalpino: Can we autodetect whether we need to pack samples or not? If so I'd prefer we did that instead of passing it in at the constructor. What happens if someone passes in packed samples and then unpacked samples after that?

src/Regressors/GradientBoost.php
```diff
@@ -536,7 +536,7 @@ public function predict(Dataset $dataset) : array
      */
     public function featureImportances() : array
     {
-        if (!isset($this->ensemble, $this->featureCount)) {
+        if (!$this->ensemble || !$this->featureCount) {
```
andrewdalpino: This new lines is not the same as the old logic. Why did you change it?

src/Regressors/Adaline.php
```diff
@@ -262,25 +263,25 @@ public function losses() : ?array
...
+     * @param Labeled $dataset
      */
     public function train(Dataset $dataset) : void
     {
         DatasetIsNotEmpty::with($dataset)->check();
 
         $this->network = new FeedForward(
             new Placeholder1D($dataset->numFeatures()),
-            [new Dense(1, $this->l2Penalty, true, new Xavier2())],
+            [new Dense(1, $this->l2Penalty, true, new XavierUniform())],
```
andrewdalpino: Since XavierUniform implements Xavier1 initialization, this change is not equivalent to the old line. We'll need to implement Xavier2Uniform and rename XavierUniform to Xavier1Uniform.

src/Classifiers/SoftmaxClassifier.php
```diff
@@ -288,7 +288,7 @@ public function train(Dataset $dataset) : void
 
         $this->network = new FeedForward(
             new Placeholder1D($dataset->numFeatures()),
-            [new Dense(count($classes), $this->l2Penalty, true, new Xavier1())],
+            [new Dense(count($classes), $this->l2Penalty, true, new XavierNormal())],
```
andrewdalpino: XavierNormal is not equivalent to Xavier1 which samples from a uniform distribution.

src/Classifiers/MultilayerPerceptron.php
```diff
@@ -369,7 +369,7 @@ public function train(Dataset $dataset) : void
 
         $hiddenLayers = $this->hiddenLayers;
 
-        $hiddenLayers[] = new Dense(count($classes), 0.0, true, new Xavier1());
+        $hiddenLayers[] = new Dense(count($classes), 0.0, true, new XavierNormal());
```
andrewdalpino: XavierNormal is not equivalent to Xavier1 which samples from a uniform distribution.

src/Classifiers/LogisticRegression.php
```diff
@@ -292,7 +292,7 @@ public function train(Dataset $dataset) : void
 
         $this->network = new FeedForward(
             new Placeholder1D($dataset->numFeatures()),
-            [new Dense(1, $this->l2Penalty, true, new Xavier1())],
+            [new Dense(1, $this->l2Penalty, true, new XavierNormal())],
```
andrewdalpino: XavierNormal is not equivalent to Xavier1 which samples from a uniform distribution.

src/Regressors/GradientBoost.php
```diff
@@ -511,7 +511,7 @@ public function train(Dataset $dataset) : void
      */
     public function predict(Dataset $dataset) : array
     {
-        if (!isset($this->ensemble, $this->featureCount, $this->mu)) {
+        if (!$this->ensemble || !$this->featureCount || !$this->mu) {
```
andrewdalpino: This new line is not the same as the old line. Why did you change it?

phpstan-bootstrap.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+// NumPower dtype names are used as constants across the codebase.
+// During static analysis these constants may not exist, so we define them here
+// to prevent undefined constant errors in PHPStan.
```
andrewdalpino, line 7: Thanks for the comment!

src/Regressors/GradientBoost.php
apphp: My mistake - turned back.

src/Regressors/GradientBoost.php
apphp: Reverted to original.

src/Classifiers/SoftmaxClassifier.php
apphp: Fixed with Xavier1Uniform

src/Classifiers/MultilayerPerceptron.php
apphp: Fixed with using Xavier1Uniform

src/Classifiers/LogisticRegression.php
apphp: Fixed with using Xavier1Uniform

src/Regressors/Adaline.php
apphp: Fixed with Xavier2Uniform

src/Regressors/MLPRegressor.php
apphp: Refactored with auto-detection of arrays, that need pack operation - see array_pack() function

andrewdalpino: Sweet! Nice work man

src/NeuralNet/Initializers/Xavier2Uniform.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * @inheritdoc
+     */
+    public function initialize(int $fanIn, int $fanOut) : NDArray
+    {
+        $this->validateFanInFanOut(fanIn: $fanIn, fanOut: $fanOut);
+
+        // Xavier-2 uses fourth-root scaling instead of standard square-root Xavier 1 scaling.
+        $limit = (6.0 / ($fanOut + $fanIn)) ** 0.25;
```
andrewdalpino, line 37: Nice, yeah Xavier2 uses that fourth root instead of square root.

andrewdalpino (dismissed):
a

src/functions.php
```diff
@@ -246,4 +247,40 @@ function warn_deprecated(string $message) : void
...
+     * Pack an array of samples.
+     *
+     * @internal
+     *
+     * @param array<mixed> $samples
+     * @param int $depth
+     * @param int $maxDepth
+     * @return array<mixed>
+     */
+    function array_pack(array $samples, int $depth = 0, int $maxDepth = 100) : array
```
andrewdalpino: Do we need the $depth and $maxDepth arguments?

> "PHP arrays are finite trees (no circular references), so infinite recursion can't happen."

Do we need these arguments for anything beyond preventing infinite recursion? If not, then let's simplify the API. Example (untested) ...

```php
function array_pack(array $samples): array
{
        if (!array_is_list($samples)) {
            $samples = array_values($samples);
        }
 
        return array_map(function ($item) {
            if (is_array($item)) {
                if (!array_is_list($item)) {
                    $item = array_values($item);
                }

                return array_pack($item);
            }
            
            return $item;
        }, $samples;
}
```

andrewdalpino (commented):
Ok Xavier2 Uniform looks good. Did you want to add Xavier2 Normal just to match the Xavier1 Normal you added?

Nice work regarding array_pack, I like the idea of building the optimization into the function itself. It handles the case of a semi-packed array, an edge case, but bound to happen to someone - so this approach solves that.

I'm asking about the need for depth arguments in the comments. Let me know what you think.

src/functions.php
apphp: Done. It was only for preventing infinite recursion

apphp:
> Ok Xavier2 Uniform looks good. Did you want to add Xavier2 Normal just to match the Xavier1 Normal you added?

Added, you may check.

src/functions.php
```diff
@@ -246,4 +247,32 @@ function warn_deprecated(string $message) : void
...
+    function array_pack(array $samples) : array
+    {
+        // Ensure all levels have sequential numeric keys
+        if (!array_is_list($samples)) {
+            $samples = array_values($samples);
+        }
+
+        return array_map(function ($item) {
+            if (is_array($item)) {
+                if (!array_is_list($item)) {
```
andrewdalpino: It looks like you copied my example, that's perfectly ok, but I didn't spend any time optimizing it. Could this check and pack block here be removed since it will be called first thing when the function recurses?

apphp: Yes, you're right - it's redundant. I removed it.

andrewdalpino (approved):

PR #418, opened 2026-07-25
417 convert classifiers to numpower

apphp:


src/NeuralNet/ActivationFunctions/Softmax.php
```diff
@@ -26,37 +28,34 @@ class Softmax implements ActivationFunction, OBufferDerivative
...
+            foreach ($values as $row) {
+                $maximum = max($maximum, $row[$column]);
             }
 
-            $result[] = $softmaxRow;
+            $maxima[] = $maximum;
         }
 
-        return NumPower::array($result);
+        $max = NumPower::reshape(NumPower::array($maxima), [1, $columns]);
```
andrewdalpino, line 54: Did we forget to implement this during the Activation Functions sprint? Was it intentional to leave out due to the NumPower max() API limitation?

src/NeuralNet/Layers/Multiclass.php
```diff
@@ -158,16 +158,17 @@ public function back(array $labels, Optimizer $optimizer) : array
                 . ' before backpropagating.');
         }
 
+        // Build one-hot targets as [classes, batch] to match Dense output layout.
         $expected = [];
 
-        foreach ($labels as $label) {
-            $dist = [];
```
andrewdalpino, line 164: What was the purpose of this change?

andrewdalpino (commented):
Looks good @samuel-akopyan nice work as usual. I a have a couple questions and things that stood out to me.

src/NeuralNet/ActivationFunctions/Softmax.php
apphp: Yes, it was Intentional. Softmax was rewritten to NumPower with stable max-subtraction, because NumPower::max() has no axis (stubs: scalar-only). Once NumPower adds max(..., axis:), we can drop the PHP loop.

src/NeuralNet/Layers/Multiclass.php
apphp: Dense outputs [classes, batch], but the old one-hot was [batch, classes]. That shape mismatch broke SoftmaxClassifier training. This change should build the targets in the same layout as the network output.

andrewdalpino (approved):
Looks good @apphp!

PR #419, opened 2026-08-15
Fix KNN Regressor weighted prediction

andrewdalpino:
src/Regressors/KNNRegressor.php:211-224 (root cause :232-246) — HIGH In weighted mode, predictSample builds $weights in distance order but array_values($labels) in original-index order, and Stats::weightedMean pairs them positionally — so each label is multiplied by the weight of a different neighbor.

Verified: train [[1],[2],[3]], labels [0,10,30], k=3 weighted, predict 2.0 → returns 10.0, correct is 12.5.
Verified: [[1],[5],[10]], [0,5,100], k=3 weighted, predict 10 → 8.55, correct is 79.60.
The classifier sibling KNearestNeighbors.php:229-233 does this correctly (aligns by key: $weights[$labels[$i]] += ...).
Fix: build weights keyed by neighbor index and pass $labels + $weights un-reordered.


apphp (approved):

apphp:
LGTM

PR #420, opened 2026-08-15
Fix Huber Loss gradient scaling

andrewdalpino:
src/NeuralNet/CostFunctions/HuberLoss.php:110 — differentiate returns e/√(e²+α²) while compute implements α²(√(1+(e/α)²)−1) whose true gradient is α·e/√(α²+e²) (missing factor α). Constant scale error; direction correct.


apphp (approved):

apphp:
LGTM

andrewdalpino:
Should probably change the variable name of the local $alpha variable so it doesn't clash with the alpha hyper-parameter.

PR #421, opened 2026-08-15
Fix TSNE gradient computation

andrewdalpino:
t-SNE gradient is mathematically wrong — src/Transformers/TSNE.php:523-543 The q-kernel uses raw (non-squared) Euclidean distances (Euclidean::compute() returns sqrt(Σ(a−b)²)), the gradient weight is raw d instead of (1 + d²/ν)^⁻¹, and Q is normalized by sum()×2 including the diagonal (all 1.0). The embedding does not minimize the documented KL divergence; wrong output is produced silently.


apphp (approved):

apphp:
Very good fix!

PR #422, opened 2026-08-15
Fix KDTree Edge pruning + optimize traversal

andrewdalpino:
KDTree nearest()/range() can return wrong neighbors — src/Graph/Trees/KDTree.php:202-211, 267-278, src/Graph/Nodes/Box.php:148-152 Pruning tests use corner distances (Box::sides() yields only min/max corners). A corner distance is neither a valid lower nor upper bound for points inside the box, so valid neighbors get pruned. Reproduced: nearest() missed true neighbors in ~5/200 trials; range() returned 0 points where brute force found 4–7. Affects KNN, RadiusNeighbors, KDNeighbors, DBSCAN, MeanShift when KDTree is selected (BallTree/VantageTree are correct).


apphp (changes requested):
@andrewdalpino please resolve merge conflict

andrewdalpino:
All set @apphp

apphp (approved):

PR #423, opened 2026-08-16
Fix Batch Norm

andrewdalpino:
BatchNorm normalizes by width, not batch size — src/NeuralNet/Layers/BatchNorm.php:184, 290-293. Matrix::variance() divides by row count (features); backward uses dOut->m() where batch size n() belongs. Masked by square test matrices.


apphp (approved):

apphp:
LGTM

PR #424, opened 2026-08-16
Fix cross entropy loss compute()

andrewdalpino:
CrossEntropy compute() missing the (1−t)log(1−p) term — src/NeuralNet/CostFunctions/CrossEntropy.php:34-39. Loss is always 0 for target=0 samples (gradient differentiate is correct BCE, so they disagree). Verified.


apphp (approved):

andrewdalpino:
Going to close this one @apphp because Co Pilot is correct, this will effect how cross entropy is calculated for multiclass (non-binary) classification problems as well. Perhaps we should consider implementing both Binary and Multiclass cross entropy loss functions in 3.0. Thoughts?

PR #431, opened 2026-08-22
Optimize Dataset sort()

andrewdalpino:
Dataset::sort() is an O(n²) bubble sort — confirmed at src/Datasets/Dataset.php:441-459. Switch to native usort() (Timsort, O(n log n), C-implemented). 10–100× for n > ~1000.

| n | Before (bubble, O(n²)) | After (usort, O(n log n)) | Speedup |
| -- | -- | -- | -- |
| 100 | 0.765 ms ±1.2% | 0.384 ms ±1.3% | 2.0× |
| 1,000 | 44.343 ms ±0.7% | 2.961 ms ±1.0% | 15.0× |
| 2,500 | 195.562 ms ±1.0% | 8.288 ms ±1.2% | 23.6× |
| 5,000 | 759.693 ms ±0.7% | 17.776 ms ±1.2% | 42.7× |

Targetting 3.0 branch as this changes the operation from an exact sort to an unstable sort.


apphp:
@andrewdalpino do we need this file in repo: runtime/test-results ?
What do you thing if we add it to .gitignore?

apphp (approved):

andrewdalpino:
> @andrewdalpino do we need this file in repo: runtime/test-results ? What do you thing if we add it to .gitignore?

Good catch @apphp, no we don't need it, I saw you added it to .gitignore, not sure why it didn't exclude it when I `git add .`. Maybe we need to specify a wildcard for subfolders?

PR #447, opened 2026-08-26
TSNE symmetrize affinity matrix

andrewdalpino:
affinities() produced a row-normalized, asymmetric P matrix (each row sums to 1, total sum = n). But gradient() uses the KL divergence gradient formula that's only valid when P is symmetric and globally normalized (total sum = 1). The algorithm was converging to the local minimum of the wrong objective — confirmed by relative error ≈ 1.6–2.2 in gradient verification.


andrewdalpino:
This exists on master as well, redoing with master as target

PR #449, opened 2026-08-27
Fix SVC Save/Load missing class map

andrewdalpino:
The SVM extension does not implement Serializable and is abandoned. So we have to work with the poor save/load mechanism.

SVC::load() drops the label map (src/Classifiers/SVC.php:272)

```php
public function load(string $path) : void {
    $this->model = new svmmodel($path);   // $classes never restored
}
```

$classes defaults to [] (line 72), is only set in train() (line 200), and predictSample() does $this->classes[$index] (line 249). After a save/load cycle, predict() indexes an empty array → ""/warning. Fix: persist & restore the class map (or rebuild from the model's labels).


torchello:
Another review from Fable. Hope it helps.

----------

# Review: RubixML/ML#449 — Fix SVC Save/Load missing class map

**Verdict:** the core fix is legitimate — it repairs a real bug (the class map lost across `save()`/`load()`), and the reworked test genuinely exercises it (fresh-instance load). No "reject" blockers. However, the new persistence path introduces **two silent-mislabel state bugs**, a wrong `@throws` contract added by the PR itself, and an unlabeled BC break with a circular remediation message.

**Process:** 10 independent finder angles → ~40 raw candidates deduped to 15 → one verifier per candidate (many verified empirically on PHP 8.5: the `json_encode`-false path, the reader/writer race, all five malformed-sidecar shapes) → final gap sweep. One candidate refuted. phpstan level 8 (CI's exact command), php-cs-fixer (0/652), CHANGELOG placement, mkdocs admonitions, benchmarks — all clean.

---

## Correctness

### 1. [HIGH] Torn state in `load()`: `$classes` assigned before the `svmmodel` constructor — `src/Classifiers/SVC.php:314-315` · CONFIRMED

The sidecar checks pass, line 314 overwrites `$this->classes`, then `new svmmodel($path)` throws (corrupt/missing model file — php-svm does `SVM_THROW("Failed to load the model")`). No rollback: a trained estimator is left with its **old model paired with the new class map**. A caller that catches the exception and keeps using the instance gets model A's indices resolved through B's labels — "spam" instead of "cat" — silently (the isset guard passes while the counts overlap). Pre-PR `load()` was a single atomic assignment.

**Fix:** decode into locals, construct the `svmmodel` first, assign both fields only after success.

### 2. [HIGH] Non-atomic `save()`: model written before sidecar → stale pair on disk — `SVC.php:276-284` · CONFIRMED

Overwrite-save to a path holding an existing pair: `$this->model->save()` has already replaced the model, then `file_put_contents` fails (empirically verified: on EACCES it returns `false` **without** touching the old file), or the process dies between the two writes (no exception at all). Disk now holds the new model + the old, still-valid sidecar; the next `load()` passes every check and silently returns the previous training run's labels ("Unknown class index" only for indices beyond the stale map). Truncation-style failures fail loudly; the intact-stale-sidecar cases are silent.

**Fix:** write to a temp file + `rename()`.

### 3. [MED] `json_encode === false` written as a 0-byte sidecar with a misleading error — `SVC.php:280-284` · CONFIRMED

A label containing invalid UTF-8 (Latin-1 bytes from a CSV, e.g. `"caf\xE9"` — verified on PHP 8.5): `json_encode` returns `false`, `file_put_contents` coerces it to `''` — **truncating a previously valid sidecar to 0 bytes** — returns `int(0)`, and `!$success` throws "Could not save the class map", blaming the filesystem when the real cause is encoding, after the model file was already overwritten. Pre-PR such labels saved fine. Bonus: `!$success` cannot detect a short write (N>0 bytes on ENOSPC) — it only surfaces later at `load()` as "malformed".

**Fix:** `Rubix\ML\Helpers\JSON::encode()` — throws `JSONException` (extends the repo's `RuntimeException`, so the documented contract is preserved) with the precise cause, before any write.

### 4. [MED] BC break: every previously saved model is permanently unloadable, and the error's advice is circular — `SVC.php:297-300` + CHANGELOG · CONFIRMED

Models saved by ≤2.5.8 have no sidecar → `load()` now always throws. The advice "re-save the model with the current version" is unactionable: `save()` requires a trained estimator, the only restore path is the very `load()` that throws, `$classes` has no setter, and the string labels are unrecoverable from the libsvm file (`train()` feeds it only flipped integer indices). The working rescue — hand-writing `<path>.classes.json` (it passes validation) — is stated nowhere; the CHANGELOG presents this as a plain fix in the 2.5.9 patch section with no breaking-change note. Calibration: pre-PR, predictions from a cross-process load already died with a TypeError, so no old artifact truly worked — the defect is the impossible advice plus the unlabeled format break, not lost functionality.

### 5. [MED] The PR-added `@throws RuntimeException` is wrong: `svmexception` extends bare `\Exception` — `SVC.php:291,315` · CONFIRMED

Verified in the extension's C source: `svmexception` is registered from `zend_exception_get_default()` (base `\Exception`, and `final`) in every supported extension version, including pecl 0.2.0 — the exact floor `ExtensionMinimumVersion('svm', '0.2.0')` enforces. The most common failure — sidecar present, model file corrupt (partial deploy) — escapes a `catch (RuntimeException)` written to the PR's own docblock, and does so **after** `$classes` was already mutated (finding 1). Same leak in `save()` via `$this->model->save()`.

**Fix:** wrap the svmmodel calls and rethrow as Rubix `RuntimeException` — one hierarchy for both halves of a load failure.

### 6. [MED] Zero sidecar↔model binding: a foreign map with a matching class count silently relabels every prediction — `SVC.php:293-316` · CONFIRMED

`load()` accepts any well-formed JSON array next to any loadable model. A ham/spam model beside a sidecar copied from a cat/dog model (or the residue of finding 2): every sample comes back "cat"/"dog" with no error anywhere. Cheap to close: php-svm exposes `svmmodel::getNrClass()`/`getLabels()` (verified registered in the C source), and the libsvm model file is plain text with `nr_class` / `label 0 1 ...` header lines — a few lines of parsing give `count($classes) === nr_class` plus "labels are exactly 0..k-1". Catches count mismatches (same-count relabelings need a checksum, the pattern `Serializers\RBX` already uses).

### 7. [MED] The sidecar schema encodes array *positions*, but `predict()` returns the raw libsvm *label* — `SVC.php:255-261` · PLAUSIBLE

`svm_predict` → `RETURN_DOUBLE(predict_label)` — the label as stored in the model, not an index. Position == label holds only for models trained by `SVC::train()` (which feeds 0..k-1). `load()` is a public API for any libsvm file: a CLI-trained model with conventional +1/−1 labels and a hand-authored sidecar `["neg","pos"]` → "Unknown class index: -1" for every negative sample; a {1,2}-labeled model silently returns the **second** sidecar entry for label-1 samples (off-by-one mislabel) and throws for label 2. Either map through `getLabels()`, or document that `load()` accepts only `SVC::save()` output.

### 8. [MED-LOW] Sidecar validation is `is_array` only; all five malformed shapes pass — `SVC.php:308-314` · CONFIRMED (all five verified empirically)

`{"1":"female","0":"male"}` → `array_values` yields `['female','male']` — mapping inverted versus the keys, no error; `[1,2]` → predictions become `"1"`/`"2"` via weak-mode `: string` coercion; `[["a"],"b"]` → TypeError from `predictSample` at predict time, far from `load()`; `[]` → every predict throws; `["male", null]` → `isset` (vs `array_key_exists`) misreports "Unknown class index: 1" for an existing-but-null entry. The gate exists precisely to vet foreign files — and accepts them all.

**Fix:** require a non-empty list of strings (`$classes === array_values($classes)` is PHP 7.4-compatible).

### 9. [MED-LOW] The sidecar filename is stated nowhere user-visible; the load() docs don't know the new contract — `docs/classifiers/svc.md:36-43` · CONFIRMED

The note exists only under `save()`, and even it never names `<path>.classes.json`; the `load()` section is a bare signature — no required file, no exception, no "the two files must travel together". A `COPY svc.model` / `aws s3 cp` / `scp` ships exactly the file passed to `save()`: the sidecar silently isn't shipped and production `load()` throws — the PR recreates its own failure mode for freshly saved models.

### 10. [LOW] Writer takes LOCK_EX, reader takes no LOCK_SH → transient spurious "malformed" — `SVC.php:280 vs 302` · CONFIRMED (race reproduced: reader got the partial prefix `["cat,"d`)

Only multi-process hot-reload of the same path; loud and fail-safe, never silent. The temp+rename from finding 2 closes this too.

### 11. [LOW] `save()` writes to a derived path with no guard: `save('foo.classes.json')` clobbers model `foo`'s sidecar — `SVC.php:278` · PLAUSIBLE

The library silently writes a file the caller never named. The trigger is contrived, but the reserved suffix is neither documented nor protected.

---

## Cleanup / architecture / tests

### 12. [altitude] A third persistence format, forking the previously byte-identical SVC/SVR/OneClassSVM save/load family — `SVC.php:270-316` · CONFIRMED

Raw libsvm file + unversioned loose JSON, versus the framework's existing depth (Persistable + `Serializers\RBX`: magic number, VERSION check, crc32b checksums, `ClassRevisionMismatch`). Confirmed rider: hyper-params/`$svm` are not restored on load — `params()`/`__toString()` misreport, a subsequent `train()` runs with mismatched options, and the PR's test masks it by rebuilding the fresh SVC with setUp-identical args. Deeper option (to propose, not demand): make SVC `Persistable` via `__serialize` embedding the model bytes + classes + params.

### 13. [reuse] `SVC.php:280,308` are now the only raw `json_encode`/`json_decode` calls in all of `src/` · CONFIRMED

Everything else (Report, RBX, NDJSON) goes through `Helpers\JSON` with precise diagnostics; the file checks duplicate `Persisters\Filesystem` minus its distinct missing-vs-unreadable messages and empty-payload guards. `JSON::encode`/`decode` are drop-in (signature- and exception-compatible). Rider: the new `use function` block imports 8 primitives while `array_flip`/`array_merge`/`array_map` in the same file stay unimported, and the `svm`/`svmmodel` class imports get split below the block.

### 14. [test] `loadWithoutClassMap`: the 512-sample training is dead weight (~99% of the test's wall time), and the assertion pins only the generic exception class — `tests/Classifiers/SVCTest.php:194-208` · CONFIRMED

`load()` throws at the sidecar check **before** ever opening the model file, so the training adds nothing; `RuntimeException` has four distinct throw sites in `load()` alone, so the test can't distinguish its named scenario from "no files at all" and survives any reorder. The `if (file_exists(...))` guard around the unlink is dead (`save()` either wrote the sidecar or threw). Cheaper faithful form: `generate(8)` + `expectExceptionMessage('class label map')`. Riders in the happy-path test: `assertIsString`/`assertNotSame('')` are vacuous next to strict `assertContains`; `new SVC(1.0, new RBF(), true, 1e-3)` is byte-for-byte the defaults — `new SVC()` reads cleaner.

### 15. [test-coverage] The only reload-over-trained-estimator coverage was deleted — `tests/Classifiers/SVCTest.php:170` · CONFIRMED

Both tests now load into fresh instances — while the torn state of finding 1 lives exactly on the "load into an already-trained instance" path. Switching the main test to a fresh instance was correct (the old in-place load masked the original bug — leftover trained state satisfied predict); the ask is an **additional** reload-over-trained test, not a revert.

---

## Refuted / verified clean

- **Refuted:** "unchecked bool return of `$this->model->save()`" — the extension itself throws (`SVM_THROW("Failed to save the model")`) in every supported version (verified down to tag 0.2.0); execution never reaches the sidecar write on failure.
- **Clean:** phpstan level 8 (CI's exact command), php-cs-fixer (0/652 files), CHANGELOG section placement, mkdocs `admonition` support for the docs note, benchmarks (never touch save/load), save-after-load round-trip (classes repopulated), test ordering/teardown symmetry, `or` vs `||` (house style, no fixer rule), Windows path handling.

## Minimal fix package (closes 1–6 and 10)

1. Decode into locals; assign `$this->classes`/`$this->model` only after a successful `svmmodel` construction.
2. Write the sidecar to a temp file + `rename()` (also removes the reader race).
3. `Helpers\JSON::encode`/`decode` instead of raw calls.
4. Cross-check `count($classes)` against the model's `nr_class` (libsvm text header or `getNrClass()`).
5. Wrap svmmodel calls, rethrow as Rubix `RuntimeException`; fix the `@throws` docblocks.
6. Make the missing-sidecar message name the real remedies (retrain, or hand-write `<path>.classes.json`), and add one sentence naming the file under both save() and load() docs.

PR #458, opened 2026-08-29
Optimize K-means and FCM

andrewdalpino:
This optimization fuses operations in Kmeans and FCM, eliminating multiple iterations over the samples.

| Estimator | Before | After | Speedup |
| -- | -- | -- | -- |
| FuzzyCMeans | 0.118s / 12.0mb | 0.083s / 8.3mb | ~30% |
| KMeans | 0.866s | 0.777s | ~10% |

### DIFF
--- a/src/Clusterers/FuzzyCMeans.php
+++ b/src/Clusterers/FuzzyCMeans.php
@@ -279,7 +279,14 @@ class FuzzyCMeans implements Estimator, Learner, Probabilistic, Verbose, Persist
         $prevLoss = INF;
 
         for ($epoch = 1; $epoch <= $this->epochs; ++$epoch) {
-            $distances = $memberships = [];
+            $sums = $totals = [];
+
+            foreach ($this->centroids as $cluster => $centroid) {
+                $sums[$cluster] = array_fill(0, $numFeatures, 0.0);
+                $totals[$cluster] = 0.0;
+            }
+
+            $loss = 0.0;
 
             foreach ($dataset->samples() as $sample) {
                 $row = [];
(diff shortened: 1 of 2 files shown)


src/Clusterers/FuzzyCMeans.php
```diff
@@ -288,12 +295,30 @@ public function train(Dataset $dataset) : void
...
+                $sigma = 0.0;
 
-                $memberships[] = $this->membershipsFromDistances($row);
-            }
+                foreach ($row as $cluster => $distance) {
+                    $weights[$cluster] = $distance ** -$this->rho;
+                    $sigma += $weights[$cluster];
+                }
+
+                $invSigma = 1.0 / $sigma;
```
apphp, line 306: May it's better to protect this row with something like: 
`$invSigma = 1.0 / max($sigma, EPSILON);`

apphp (changes requested):

src/Clusterers/FuzzyCMeans.php
andrewdalpino: Great suggestion

andrewdalpino: I think we're safe. Here's the chain:

Line 295: each distance is floored via $distance ?: EPSILON (EPSILON = 1e‑8), so no distance is 0 → no 0 ** -rho → no INF weights.
Line 142: fuzz is validated > 1.0, so $rho = 2.0 / ($fuzz - 1.0) (line 159) is always positive.
Line 302: $distance ** -rho is therefore a positive value for every weight, and $sigma is a sum of c >= 1 positive terms, so $sigma > 0 in all realistic cases.

The only way $sigma could hit 0 is a theoretical float underflow — every distance would have to be on the order of 1e154+ so that distance ** -rho rounds to 0.0. That's not a realistic data scenario.

samuel-akopyan (approved):

samuel-akopyan (approved):

apphp (approved):

PR #468, opened 2026-09-01
Numpower data types

andrewdalpino:
Allow the user to change the data type of the underlying NDArrays. Note that due to the poor implementation of data types in the current version of NumPower we're restricting the datatype to only float32 until I can fix the implementation myself.


andrewdalpino:
I made this before I discovered all the issues with the current NumPower datatype implementation.

PR #471, opened 2026-09-03
Overall persistence improvements

andrewdalpino:
Make Filesystem writes atomic, detect cycles in object-property graph during class revision hashing, validate RBX header schema, class revision mismatch is now a warning.


src/Persisters/Filesystem.php
```diff
@@ -105,10 +122,51 @@ public function save(Encoding $encoding) : void
...
+
+            if (!fclose($handle)) {
+                throw new RuntimeException("Could not finalize the write to temp file {$temp}.");
+            }
+
+            if (!rename($temp, $this->path)) {
+                throw new RuntimeException('Could not finalize the save at ' . $this->path . '.');
+            }
+
+        } finally {
```
apphp, line 169: If writing the temp file fails, the file handle stays open and then `finally` tries to delete the temp file. That leaks the handle. On Windows, deleting a file that is still open often fails, so the temp file can be left behind.

Please close the handle in `finally` (if it is still open), then delete the temp file only if it is still there after a failed rename.

src/Serializers/RBXV2.php
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * Collect the set of class names present in an object's property graph.
+     *
+     * @internal
+     *
+     * @param object $root
+     * @return Set
+     */
+    protected static function collectClassSet(object $root) : Set
```
apphp: This walk only looks one level into arrays, and it misses private properties from parent classes.
Do you think it critical or not?

tests/Serializers/RBXV2Test.php
```diff
@@ -0,0 +1,221 @@
...
+        string $library = '3'
+    ) : string {
+        $header = $this->makeHeader($name, $revision, $set, $payload, $library);
+        $checksum = 'sha256:' . hash('sha256', $header);
+
+        return self::IDENTIFIER . $version . "\n" . $checksum . "\n" . $header . "\n" . $payload;
+    }
+}
+
+class Gadget
```
apphp: Gadget and RBXV2Carrier not in use, do you need these classes?

apphp (changes requested):

src/Persisters/Filesystem.php
andrewdalpino: Fixed, great find!

src/Serializers/RBXV2.php
andrewdalpino: It's a bug for sure, not sure if it's critical. I'm going to fix it regardless though, nice catch.

tests/Serializers/RBXV2Test.php
andrewdalpino: Nope dropping

samuel-akopyan (approved):

PR #480, opened 2026-09-08
Fix Multiclass layer gradient

andrewdalpino:
Softmax activation function is incompatible with any non-Cross Entropy-based loss function. Remove Softmax activation function. Subsume functionality into Multiclass layer - correct the gradient computation for non-Cross Entropy-based loss function such as Relative Entropy.


apphp (approved):

andrewdalpino:
Thank you @apphp!

PR #489, opened 2026-09-11
Reorder train/validation step in GBM learners

andrewdalpino:
Off-by-one in early-stopping validation. The score at epoch t is computed from $outTest before epoch t's booster is added, so it reflects the model with t−1 boosters. But bestEpoch = $epoch and restoration does array_slice($this->ensemble, 0, $bestEpoch) — so it restores one booster too many, and the reported per-epoch score is mislabeled.


andrewdalpino:
Will address this later

